PR #51, opened 2026-05-20
Replace Claude Code runtime with OpenCode

cliftonc:
## Summary

- Replace both agent runtimes (sandbox workflow phases + chat) with [OpenCode](https://github.com/sst/opencode); no `claude` binary, no `@anthropic-ai/claude-agent-sdk` import, no `--output-format stream-json` parser remains in `src/`.
- Provider-agnostic via `OPENCODE_MODEL` (default `openai/gpt-5.3-codex`); auth via `OPENAI_API_KEY` and/or `ANTHROPIC_API_KEY`.
- Dashboard contract preserved via a Claude-SDK envelope jsonl shim (`src/engine/opencode-shim.ts`); `SessionReader` and `ChatSessionReader` work unchanged.
- Two security fixes shipped alongside: harness no longer touches `~/.gitconfig` by default, and the chat agent is now tool-fenced (not just prompt-fenced) against `bash` / `edit` / destructive `github_*` tools.

Full phase-by-phase breakdown in `~/.claude/plans/we-have-a-doc-pure-curry.md` (commit history mirrors it).

## Notable changes

- **Sandbox runtime** (`src/engine/opencode-executor.ts`, `src/sandbox/docker.ts`, `sandbox.Dockerfile`): each workflow phase invokes `opencode run --format json` inside a Docker container; harness parses the event stream into `ExecutionResult`. Pinned `opencode-ai@1.15.5`.
- **Chat runtime** (`src/engine/opencode-chat-server.ts`, `src/engine/chat.ts`): long-lived `opencode serve` HTTP server, one OpenCode session per Slack thread. Per-session in-flight chain serialises same-`sessionId` calls so two quick messages in one thread can't race on server-side conversation state — different sessions stay parallel.
- **Dashboard shim** (`src/engine/opencode-shim.ts`): translates OpenCode `text` / `tool_use` / `error` events into Claude-SDK envelope jsonl under `$STATE_DIR/opencode-home/projects/<slug>/<sessionId>.jsonl`. Prefixes MCP tool names (`<server>_<tool>` → `mcp_<server>_<tool>`) so the dashboard classifier keeps routing them correctly.
- **Skills** rewritten to use OpenCode's `github_<tool>` MCP names instead of the Claude `mcp_github_<tool>` convention (24 files touched across `skills/`, `workflows/prompts/`, `agent-context/`).
- **Rate-limits feature deleted** entirely — was tightly coupled to Anthropic-specific stream-json `rate_limit_event`. Cron, admin routes, two SQLite tables, the dashboard footer, and dependent code all removed.
- **Dependency cleanup**: `npm uninstall @anthropic-ai/claude-agent-sdk` (-303 lines of lockfile). The two remaining tiny LLM call sites (`screen.ts`, `classifier.ts`) moved to a direct-`fetch` helper in `src/engine/llm.ts` that routes to Anthropic Messages or OpenAI Chat Completions by model-id prefix.
- **Env rename**: `CLAUDE_MODEL` → `OPENCODE_MODEL`, `CLAUDE_MODELS` → `OPENCODE_MODELS`, `CLAUDE_HOME_DIR` → `OPENCODE_HOME_DIR`. State dir `claude-home/` → `opencode-home/`. Clean break — no compat aliases (this is a fork branch).
- **Security:**
  - `~/.gitconfig` writes are now opt-in (`LASTLIGHT_WRITE_GLOBAL_GIT=1`); default is do-nothing. Previously any code path that didn't set `LASTLIGHT_LOCAL_DEV=1` overwrote the user's global git identity and credential helper.
  - Custom `chat` agent defined in `opencode.json` denies `bash`, `edit`, `webfetch`, `websearch`, `task`, `skill`, `todowrite`, `repo_clone`, `repo_overview`, `external_directory`, plus the destructive `github_*` write tools (`clone_repo`, `create_branch`, `push_files`, `create_or_update_file`, `*_git_auth`, `merge_pull_request`, `create_pull_request`, `create_pull_request_review`). Reads + issue/comment/label management still allowed.
- **Docs**: `CLAUDE.md`, `README.md`, `src/workflows/CLAUDE.md` rewritten for the OpenCode runtime. `docs/pi-migration-plan.md` marked SUPERSEDED. Phase-by-phase findings checked in at `docs/opencode-fork-phase3-results.md` and `docs/opencode-fork-phase4-results.md`.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npx vitest run` — 394 passed / 1 todo (added 26 new tests across the new modules)
- [x] `cd dashboard && npx tsc -b` clean
- [x] `docker compose build sandbox` builds clean (2.17 GB; `opencode-ai@1.15.5` baked in)
- [x] End-to-end sandbox run: `repo-health` on `cliftonc/drizzle-cube` exercised the full pipeline — sandbox creation, OpenCode stream parsed, `sessionID` captured into `executions.session_id`, Phase 2 shim wrote a 3-line envelope jsonl, error mapping (`error_api`) correct. Details in `docs/opencode-fork-phase3-results.md`.
- [x] `scripts/chat-smoke.mjs` two-turn HTTP probe against real `opencode serve` — session resume + multi-turn cache reuse confirmed (turn 2 `cache.read=2048` after the chat-agent fence; was 6144 before the fence shrunk the system prompt).
- [ ] Live qualitative comparison against the `main` baseline — blocked on funded API key. Recommend canary on prod for 24h before merge: `ssh root@85.9.213.18 /home/lastlight/deploy.sh`. Roll-back is `git revert` of the merge commit; the prior image is still in the registry.

## Migration notes for prod

- Existing prod databases still have the orphaned `rate_limits` + `system_status` tables. SQLite tolerates them; they vanish on a fresh data dir.
- `$STATE_DIR/claude-home/` (legacy) keeps its historical session jsonls. The new harness writes to `$STATE_DIR/opencode-home/`. Either rename the directory once or leave the legacy dir as a read-only archive.
- `.env` must be updated: `CLAUDE_*` → `OPENCODE_*`. Set `OPENAI_API_KEY` (and/or `ANTHROPIC_API_KEY`).


src/engine/llm.ts
```diff
@@ -0,0 +1,182 @@
...
+  try {
+    const parsed = JSON.parse(raw) as Record<string, unknown>;
+    const v = parsed?.[taskType];
+    return typeof v === "string" ? v : undefined;
+  } catch {
+    return undefined;
+  }
+}
+
+export function resolveProvider(model: string): { provider: "anthropic" | "openai"; modelId: string } {
```
cliftonc: [P2] Provider-agnostic claim is not true for helper LLM calls

The migration docs say OpenCode is provider-agnostic, but the classifier/screener helper path only supports Anthropic and OpenAI. Any `provider/model` prefix other than `anthropic` or `openai` falls through to OpenAI and sends the full model string as an OpenAI model id. OpenCode workflow phases may support more providers, but event routing and injection screening do not.

src/engine/opencode-chat-server.ts
```diff
@@ -0,0 +1,490 @@
...
+      read: "allow",
+      glob: "allow",
+      grep: "allow",
+      list: "allow",
+      // Web lookups + internal scratchpad — useful for conversational
+      // responses, no escape from the fence.
+      webfetch: "allow",
+      websearch: "allow",
+      todowrite: "allow",
+      // GitHub MCP — code/branch/PR mutation
```
cliftonc: [P2] Chat permissions drifted from allow-list to deny-list

The old chat path used an explicit `ALLOWED_MCP_TOOLS` list; this OpenCode config denies several destructive GitHub tools but allows any unlisted `github_*` tool. Today that permits `github_create_label`, and any future write tool added to `mcp-github-app` will be chat-callable by default. For the same safety constraints as the old version, this should become an explicit allow-list or deny every GitHub write not deliberately allowed.

src/engine/opencode-chat-server.ts
```diff
@@ -0,0 +1,490 @@
...
+    const file = path.join(this.cfg.workingDir, "opencode.json");
+    await fs.writeFile(file, JSON.stringify(config, null, 2));
+
+    if (this.cfg.agentMarkdown) {
+      await fs.writeFile(path.join(this.cfg.workingDir, "AGENTS.md"), this.cfg.agentMarkdown);
+    }
+  }
+
+  private async spawn(): Promise<void> {
+    const binary = this.cfg.binary || process.env.OPENCODE_BIN || "opencode";
```
cliftonc, line 267: [P1] Chat server cannot find OpenCode in production image

The production Dockerfile installs the old Claude CLI into PATH, but never installs or exposes an `opencode` binary. `opencode-ai` is only installed under `/app/node_modules/.bin`, and the container starts with `node dist/index.js`, so npm does not inject that directory into PATH. The chat supervisor then spawns plain `opencode`, meaning Slack/chat mode will fail on Docker startup unless `OPENCODE_BIN=/app/node_modules/.bin/opencode` or PATH/global install is added.

cliftonc:
[P2] Docker deploy docs omit required sandbox image build (`README.md:145-148`)

Workflow execution requires `lastlight-sandbox:latest`; `createTaskSandbox` refuses to sandbox if that image is missing. The Docker deployment instructions only build `agent`, while the sandbox service is behind the `build-only` profile. A clean production host following the docs will start the harness but fail every workflow unless it also runs `docker compose --profile build-only build sandbox` or an equivalent build step.

Posting as a PR conversation comment because this line is unchanged context in the PR diff, so GitHub will not accept an inline review comment there.

cliftonc:
[P1] Deployment still depends on Claude (`Dockerfile:14-19`)

The migrated top-level image still downloads and validates the Claude CLI during build, and still exports `CLAUDE_HOME_DIR`. That makes deploys depend on `claude.ai/install.sh` even though runtime execution has moved to OpenCode. This is a real deployment constraint regression: a Claude outage, blocked domain, or removed installer breaks the Last Light image before OpenCode ever runs.

Posting as a PR conversation comment because `Dockerfile` is not changed in this PR diff, so GitHub will not accept an inline review comment there.

PR #64, opened 2026-05-27
fix(sandbox): make Node fetch honor HTTPS_PROXY in docker sandboxes

cliftonc:
## Summary
Docker-mode workflows were erroring with \"Connection error\" because Node 22's built-in fetch (undici) doesn't honor \`HTTP_PROXY\` / \`HTTPS_PROXY\` env vars by default. The OpenAI/Anthropic/pi-ai SDKs all use fetch, so they dialed direct — and on \`sandbox-egress\` (\`internal: true\`) there's no host route, so every LLM call hung. Tinyproxy logs showed zero connections, confirming the SDK bypassed the proxy entirely.

\`HTTPS_PROXY\` does work for everything else in the sandbox (\`curl\`, \`gh\`, \`git\`, \`npm\`, \`python\`/\`pip\`) since those honor it natively. Node fetch is the lone exception.

## Fix
- \`deploy/sandbox-proxy-init.mjs\` — tiny preload that calls \`setGlobalDispatcher(new EnvHttpProxyAgent())\` when \`HTTP(S)_PROXY\` is set. No-op otherwise.
- \`sandbox.Dockerfile\` — \`npm install -g undici@^7\` and \`COPY\` the preload into the image.
- \`src/sandbox/docker.ts\` — when \`proxyHost\` is configured, also set \`NODE_OPTIONS=--import file:///app/sandbox-proxy-init.mjs\` in the sandbox env so every node process in the container preloads the dispatcher.

## Why not transparent?
A real transparent SNI proxy (HAProxy/sniproxy + DNAT) was considered and discarded for now — same security properties as the env-var path for the LLM-traffic case but materially more architecture. Tracked as a follow-up.

## Test plan
- [x] \`npx tsc --noEmit\` clean
- [x] \`npx vitest run\` — 436 + 1 todo, all green
- [ ] Manual on prod: run an issue-comment workflow, verify tinyproxy-strict logs show CONNECT to api.openai.com and the workflow completes

### DIFF
--- a/deploy/sandbox-proxy-init.mjs
+++ b/deploy/sandbox-proxy-init.mjs
@@ -1,0 +1,32 @@
+// Preloaded into the sandbox `node` process via NODE_OPTIONS='--import ...'.
+//
+// Node 22's built-in fetch (undici) does NOT honor HTTP_PROXY / HTTPS_PROXY
+// env vars by default — every HTTP client built on global fetch (the OpenAI
+// SDK, the Anthropic SDK, agentic-pi) ignores them and dials direct. In a
+// sandbox attached to an `internal: true` docker network, that means
+// every LLM call times out with a generic "Connection error" because the
+// only path off-network is via the tinyproxy sidecar.
+//
+// Register undici's EnvHttpProxyAgent as the global dispatcher so fetch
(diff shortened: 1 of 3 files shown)


cliftonc:
@last-light can you review this?

PR #66, opened 2026-05-28
fix(sandbox): replace tinyproxy with SNI-peeking firewall (nginx + coredns)

cliftonc:
## Why
The tinyproxy / `HTTPS_PROXY` approach is fundamentally fragile: clients inside the sandbox have to honour the proxy env vars, and SDKs that build their own undici dispatcher (notably OpenAI's `openai-responses` client used by agentic-pi) silently bypass them. We proved this in prod: \`node -e fetch(...)\` honoured \`NODE_USE_ENV_PROXY=1\` and reached tinyproxy, but the same container running \`agentic-pi run\` failed with \"Connection error\" and tinyproxy logged zero traffic.

## What
SNI-peeking firewall pattern, inspired by Vercel Sandbox. The sandbox sees no proxy at all — it dials real hostnames directly. The network does the rest.

\`\`\`
sandbox-egress (internal: true, 172.30.0.0/24)
  ├─ sandbox containers              docker run --dns <coredns-ip>
  ├─ coredns-strict   172.30.0.10    → returns nginx-strict for allowlist hosts, NXDOMAIN otherwise
  ├─ coredns-open     172.30.0.11    → returns nginx-open for ANY hostname (minus SSRF hard-deny)
  ├─ nginx-egress-strict 172.30.0.20 → ssl_preread; map SNI → upstream; tunnel via proxy-egress
  └─ nginx-egress-open   172.30.0.21 → tunnel any SNI

proxy-egress  (regular bridge, only the two nginx services)
\`\`\`

## Wildcards
Leading-dot entries in \`egress-allowlist.ts\` (e.g. \`.github.com\`) cover the apex plus every subdomain via nginx's native map syntax and an anchored CoreDNS template regex. Reduces the hardcoded GitHub list from five entries to one wildcard.

## Validated end-to-end on prod
Ad-hoc smoke before this PR proved the architecture works:

\`\`\`
$ docker run --network lastlight_sandbox-egress --dns <coredns-ip> sandbox-image
  agentic-pi run --model openai/gpt-5-mini ... 
errorMessage: OpenAI API error (401): Incorrect API key provided: sk-bogus
\`\`\`

That's a real HTTP 401 from OpenAI returned via the SNI gateway — proves the full pipeline (agentic-pi → fetch → coredns → nginx → tunnel → OpenAI) works with **zero env vars** in the sandbox.

## Test plan
- [x] \`npx tsc --noEmit\` clean
- [x] \`npx vitest run\` — 438 + 1 todo, all green
- [x] \`src/sandbox/egress-firewall-config.test.ts\` — config-gen behaviour
- [x] \`src/sandbox/docker-compose.test.ts\` — topology contract, static IP cross-check
- [ ] After merge + deploy: run an issue-comment workflow, confirm nginx-egress-strict logs the matching CONNECT line and the workflow completes

## What this still doesn't cover (honest list)
- **No TLS termination.** A hostname that A-records to a private IP would be tunnelled by nginx-open if we ever wanted that — but the matching strict-coredns wouldn't resolve it in the first place. Closing the open-mode gap requires Envoy + \`dynamic_forward_proxy\` with post-resolve IP checks. Documented in code + CLAUDE.md, tracked as follow-up.
- **HTTP / non-TLS:** the firewall is HTTPS-only. \`internal: true\` on sandbox-egress means there's no path off-network for plain HTTP anyway.

## Pre-deploy
Already updated \`/home/lastlight/deploy.sh\` on prod to drop the \`tinyproxy-strict\` reference from \`docker compose build\`. Once this merges, normal \`./deploy.sh\` will work.


cliftonc:
@last-light can you review this?

cliftonc:
@last-light can you review this?

cliftonc:
@last-light can you review this?

PR #105, opened 2026-06-18
Carve WorkflowRunStore with atomic lifecycle operations out of StateDb (#97)

### LINKED ISSUE #97: Carve a WorkflowRunStore with atomic lifecycle operations out of StateDb
## What to build

A deep `WorkflowRunStore` (and, alongside it, `ApprovalStore` / `ExecutionStore`) carved out of the `StateDb` god-class, exposing intention-level operations that own their multi-step mutations in one SQLite transaction.

`StateDb` is one class spanning seven unrelated concerns (executions, chat threads, workflow runs, approvals, cron overrides, workflow kill-switches, stat rollups) where nearly every method is a 1:1 SQL translation that enforces no invariant. Callers compensate by stitching read-modify-write chains. The clearest hazard is the socratic explore-reply path: it manually reconstructs `scratch.socratic.qa`, then calls resolve-reply-gate → update-scratch → update-phase → resume-run → dispatch with **no transactional boundary**. A throw mid-sequence leaves the run marked `running` but never re-dispatched, and the user's reply is silently lost.

Deliver domain operations such as:
- `appendSocraticReply(runId, reply)` — the scratch reconstruction + persistence as one atomic op.
- `resolveGateAndResume(runId, ...)` — resolve gate, append phase history, flip status, all-or-nothing.
- `finishRun(runId, status)`.

`StateDb` becomes a thin low-level SQLite handle the stores use internally, not the public surface callers reach into.

The deepening: run-lifecycle invariants and atomicity concentrate in one module (fix once, fixed everywhere); the store is testable without booting connectors, cron, or the chat runner; `index.ts` and `simple.ts` shed their ad-hoc state choreography.

## Acceptance criteria

- [ ] Run-lifecycle transitions (running → paused → resumed/finished, gate resolution, socratic-reply append) are exposed as named operations on a store, each wrapped in a single DB transaction.
- [ ] The explore-reply / approval-response flows call store operations instead of sequencing raw `StateDb` mutations; a failure partway through leaves no half-applied state.
- [ ] The store is unit-tested directly against an in-memory/temp SQLite DB, including the mid-sequence-failure rollback case.
- [ ] No behavior change observable to GitHub/Slack callers; existing state and workflow tests pass.

## Blocked by

None - can start immediately.

---

## Context from the #94 grilling (status-store consolidation)

#94 consolidates the runner onto the **executions ledger as the single source of truth** for phase status/resume, and deletes the dead `node_statuses` store. But it deliberately leaves a second status-ish store for this issue to reconcile:

- **`phase_history`** (a JSON array on the `workflow_runs` row, appended by `persistPhase`) overlaps with the `executions` table but also holds **non-agent markers** that aren't executions: `context` phases, the synthetic `waiting_approval` pause marker, and `on_success.set_phase` terminal markers. The dashboard reads it as a *fallback* status source (`dashboard/src/components/WorkflowPipeline.tsx` ~line 223), alongside the executions it treats as primary.

So after #94 there are still two overlapping stores (`executions` + `phase_history`). The WorkflowRunStore should **reconcile them into one model** — either fold the non-agent markers into the executions ledger (so it is the sole status source the dashboard derives from) or make `phase_history` a derived projection. Whatever the choice, the goal is a single authoritative status source the dashboard reads without duplication. Coordinate with the dashboard's derive-from-executions logic so status isn't computed two ways.

cliftonc:
Closes #97.

Carves the `StateDb` god-class into three per-table stores sharing one `better-sqlite3` connection, and concentrates run-lifecycle invariants in named, single-transaction operations. Follows the design **and the clarifying comment** on #97 (the "resolved via grilling" carve-out shape).

## Decomposition — 3 stores, one shared `Database`
- **`ExecutionStore`** (`src/state/execution-store.ts`) — owns *every* `executions` query (record/stats/dedup/chat-threads). Single owner per table.
- **`ApprovalStore`** (`src/state/approval-store.ts`) — `workflow_approvals` CRUD. `resolveReplyGate` is guarded by `kind='reply' AND status='pending'` and returns its row-change count.
- **`WorkflowRunStore`** (`src/state/workflow-run-store.ts`) — aggregate root, constructed with the injected `ApprovalStore`. Holds the atomic ops.
- **`StateDb`** (`src/state/db.ts`) — now a thin construction root (1428 → 210 lines): opens the connection, runs the extracted `migrate()`, builds the stores, exposes `db.runs` / `db.approvals` / `db.executions`, **re-exports the moved types** (zero import churn), keeps only cron-override + kill-switch concerns. Old methods **deleted**.

## Named atomic ops (each one `db.transaction(...)`)
- `resolveReplyGateAndResume` — the socratic explore-reply hazard the issue calls out: resolve reply gate + merge `scratch.socratic.qa` + resume, all-or-nothing. Doubles as the double-reply concurrency guard.
- `resolveGateAndResume` / `resolveGateAndFail` — approve / reject.
- `pauseForApproval` — folds approval insert + `waiting_approval` marker + optional scratch persist + pause (covers the standard gate, reviewer-loop, and generic-loop pause sites).
- `finishRun({ terminalMarker })` — folds the `on_success.set_phase` terminal marker into the status flip.

The long-running `dispatchWorkflow` (spawns a Docker sandbox for minutes) stays **outside** every transaction — the atomic boundary is the DB mutations only; callers dispatch after commit.

## Call sites
All ~40 rewired across `index.ts`, `admin/routes.ts`, `simple.ts`, `runner.ts`, `phase-executor.ts`, `resume.ts`, `router.ts`, `scheduler.ts`, `chat-session-reader.ts`. The explore-reply and both approval-response flows (GitHub/Slack **and** dashboard) now call the atomic ops — a throw mid-sequence can't leave a run `running` with its gate unresolved or its reply lost.

## Deferred (per the issue)
`phase_history`/`executions` reconciliation is out of scope to honor "no behavior change". The `appendPhase` seam is created now so that follow-up touches one method instead of ~10 call sites.

## Acceptance
- [x] Lifecycle transitions exposed as named, single-transaction ops on a store.
- [x] explore-reply / approval-response flows call store ops; partial failure leaves no half-applied state.
- [x] Store unit-tested directly incl. the mid-sequence-failure rollback case.
- [x] No behavior change observable to GitHub/Slack callers; existing state + workflow tests pass.

## Tests
New `workflow-run-store.test.ts` drives every atomic op, including **two rollback proofs** — a BigInt poison patch (natural `JSON.stringify` throw mid-txn) and an injected throwing `ApprovalStore` — plus the double-reply guard. Existing state/runner/executor/routes/router tests migrated to the store APIs.

`:memory:` is now a genuine per-connection in-memory DB so parallel test files don't contend on a shared on-disk file once the stores hold write transactions.

**`tsc --noEmit` clean · `vitest run` 601/601 pass.**

### Note
One deliberate, non-observable wording change: the unified `resolveGateAndFail` sets the run's internal `context.error` to `Rejected: <reason>` for both channels (was `Rejected by <sender>: …` / `Rejected via dashboard: …`). The responder is still recorded on the approval row (`respondedBy`); only the internal error annotation converges.


cliftonc:
@last-light lets fix this

cliftonc:
Fixed in 24bcca3. The dashboard approve path no longer pre-flips the run via the atomic `resolveGateAndResume`. It now records the approval response, then delegates the status flip to `resumeWorkflow`, which already validates the dispatch target (GitHub App present, trigger is an `owner/repo#N` issue) and calls `setRunning` immediately before dispatching. So a non-resumable approval — no `resumeWorkflow` wired, App down, or a non-issue trigger — is recorded but leaves the run `paused` rather than orphaning it as `running` with no worker. This restores the pre-#105 semantics.

`resolveGateAndResume` stays in use on the GitHub/Slack path (`index.ts`), where the dispatch target is validated *before* the call, so the flip is always followed by a dispatch. Updated its doc comment to spell out why the dashboard deliberately doesn't use it.

Added `POST /approvals/:id/respond` route tests covering the resumable, non-resumable (the regression), and rejection paths. Full suite green: 633 tests, `tsc` clean.

cliftonc:
Fixed in bccb6c2. The dashboard approve path now respects the compare-and-set return from `ApprovalStore.respond()`. Since `respond()` updates `WHERE id = ? AND status = 'pending'` and returns the row count, a racing responder that loses changes 0 rows — so the path guards on `changed === 1`, the loser returns 409, and `resumeWorkflow` is only handed the run on the winning response. This closes the TOCTOU window between the status read and the write. The GitHub/Slack path already inherits the same guard via `resolveGateAndResume` / `resolveGateAndFail`.

Added a CAS-race test asserting the 409 + no-resume behavior. Full suite green: 634 tests, `tsc` clean.

(FYI on the verification ENOSPC — that's the review sandbox running out of disk during `npm ci`, not something in the diff.)

PR #147, opened 2026-07-11
feat(auth): OAuth subscription logins (Codex, Claude Pro, Copilot)

alfonsograziano:
> ### ✅ Update — Codex now works in the **sandbox** too (not just chat)
>
> The original "Codex is chat-only" limitation below is **resolved**. The fix
> landed upstream in **[nearform/agentic-pi#9](https://github.com/nearform/agentic-pi/pull/9)**
> (`authFile` / `--auth-file`): since the model call runs in the agentic-pi
> process — the sandbox executes only *tools* — pointing Pi's `AuthStorage` at a
> host credential store is all a subscription login needs, no token in the VM.
>
> The follow-up commit here (`feat(oauth): Codex in the sandbox via agentic-pi authFile`)
> threads `$STATE_DIR/auth.json` to agentic-pi as `authFile` for the in-process
> backends (`none`/`gondolin`), so **Codex authenticates in sandbox workflow
> phases** there. Container backends (`docker`/`smol`) still carry Anthropic/
> Copilot via env tokens; Codex there awaits mounting the store into the guest.
> Verified end-to-end: a Codex model ran through Last Light's `executeAgent`
> sandbox path (backend `none`) with only the OAuth login — real cost, clean finish.
>
> **Merge order:** requires agentic-pi#9 to ship first; then bump the `agentic-pi`
> dependency here (CI typecheck needs the new `authFile` type).
>
> ---

## What

Adds support for authenticating the model provider with a **subscription login** (OAuth) instead of a static API key, via pi-ai's OAuth providers: **ChatGPT Plus/Pro (Codex)**, **Claude Pro/Max**, and **GitHub Copilot**.

Previously `src/providers.ts` deliberately excluded these ("OAuth-only providers") — there was no login command, credential store, or wiring.

## How

- **`src/engine/oauth.ts`** (new) — shared credential layer: one on-disk store (`$STATE_DIR/auth.json`, same JSON shape pi-ai's own CLI writes; override `LASTLIGHT_AUTH_FILE`), `resolveOAuthApiKey()` (refresh-if-expired + persist rotated creds), and the model-prefix→provider-id / sandbox-env-var maps.
- **`src/providers.ts`** — `OAUTH_PROVIDERS` registry as the single source of truth, alongside the API-key `PROVIDERS`.
- **Chat path** (`chat-runner.ts`) — resolves a token per turn and passes it as the per-call `apiKey`; an OAuth-only model with no login fails the turn with an actionable *"run: lastlight oauth login …"* error.
- **Sandbox path** (`agent-executor.ts`) — injects `ANTHROPIC_OAUTH_TOKEN` / `COPILOT_GITHUB_TOKEN` into the sandbox env.
- **CLI** — `lastlight oauth login|list|status|test|logout` (host-local; `src/cli/oauth-cli.ts`).
- **Setup wizard** — key-less OAuth provider option that points at the login.
- **Docs** — README, CLAUDE.md, `.env.example`.

## Key constraint: two seams, different reach

pi-ai exposes OAuth two ways, and they don't have equal reach:

- **Chat** (in-process pi-ai) passes the token as an `apiKey` → **all three providers work, Codex included.**
- **Sandbox** (agentic-pi) resolves creds from env only. Anthropic/Copilot work via their env tokens, but **Codex has no env-token route (chatgpt.com backend), so it's chat-only** — the executor warns instead of failing opaquely mid-run. No OAuth host is added to the sandbox egress allowlist (chat is in-process; Anthropic OAuth reuses the already-allowed `anthropic.com`).

## Testing

26 tests across the real behaviors (not just helpers):

- `tests/engine/oauth.test.ts` — store round-trip (mode 0600), auth-file resolution precedence, and registry consistency (oauth.ts's derived maps stay in sync with `OAUTH_PROVIDERS`).
- `tests/engine/chat-runner.oauth.test.ts` — the three chat decisions: OAuth-only-no-login → error + model **not** called; login → token passed as `apiKey`; API-key-capable (Anthropic) no-login → fallthrough; refresh-throws → error; plain provider → OAuth untouched.
- `tests/engine/agent-executor.oauth.test.ts` — sandbox env injection via `FakeSandbox`: Anthropic/Copilot → token env vars; Codex → nothing (never resolves); no-login → nothing; preset token → skip resolve.

Full suite: **988 passing / 7 skipped**; `typecheck:test` clean.

Verified manually end-to-end: `lastlight oauth login openai-codex` → a live chat turn served through a ChatGPT Codex subscription with no API key.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -54,6 +54,20 @@ WEBHOOK_SECRET=your-webhook-secret
 # KIMI_API_KEY=...                     # kimi-coding/… models
 # MINIMAX_API_KEY=...                  # minimax/… models
 
+# ── Subscription logins (OAuth — no API key) ───────────────────────────────
+# Instead of a provider API key you can authenticate with a subscription:
+#   ChatGPT Plus/Pro (Codex), Claude Pro/Max, or GitHub Copilot. Log in once
+#   on this host — it writes $STATE_DIR/auth.json (the same shape pi-ai's own
+#   CLI writes) which the harness reads:
+#     lastlight oauth login openai-codex     # ChatGPT Plus/Pro (Codex)
+#     lastlight oauth list                   # see providers + login status
+#   then point the model at that provider and restart:
+#     LASTLIGHT_MODEL=openai-codex/gpt-5.6-terra
+# Reach: Codex works for CHAT only (its chatgpt.com backend has no env-token
+# route, so it can't authenticate inside the agentic-pi sandbox). Anthropic
+# and Copilot OAuth also work for sandbox workflows via ANTHROPIC_OAUTH_TOKEN /
+# COPILOT_GITHUB_TOKEN. Override the store path with LASTLIGHT_AUTH_FILE.
+
 # ── Slack OAuth (optional — enables "Login with Slack" on dashboard) ──
 # Create a Slack app at https://api.slack.com/apps and enable "Sign in with Slack"
 # Redirect URI must be: https://<your-host>/admin/api/oauth/slack/callback
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -32,6 +32,22 @@ API credentials are read from the provider env vars in the registry at
 match your `OPENCODE_MODEL` / `OPENCODE_MODELS`. No `claude` CLI, no
 Anthropic SDK in the runtime path.
 
+**Subscription logins (OAuth).** Besides the API-key providers above, three
+providers authenticate by subscription login instead of a static key —
+`openai-codex` (ChatGPT Plus/Pro), `anthropic` (Claude Pro/Max), and
+`github-copilot`. They're registered in `OAUTH_PROVIDERS` in `src/providers.ts`
+(separate from the API-key `PROVIDERS`). `src/engine/oauth.ts` is the shared
+layer: one on-disk store (`$STATE_DIR/auth.json`, override `LASTLIGHT_AUTH_FILE`
+— same JSON shape pi-ai's own CLI writes), `resolveOAuthApiKey()`
+(refresh-if-expired + persist), and the model-prefix→provider-id map.
+`lastlight oauth login|list|status|test|logout` (host-local,
+`src/cli/oauth-cli.ts`) drives the browser flow. **Two seams, different reach:**
+the in-process **chat** path (`chat-runner.ts`) passes the token as a per-call
+`apiKey`, so all three work; the **sandbox** path (`agent-executor.ts`) resolves
+creds from env only and injects `ANTHROPIC_OAUTH_TOKEN` / `COPILOT_GITHUB_TOKEN`
+— Codex has no env-token route, so it's **chat-only** (the executor warns if a
+Codex model is used for a sandbox phase).
+
 The cheap-helper path (`src/engine/llm.ts`, used by screener + classifier)
 bypasses OpenCode and dispatches directly to the same three providers.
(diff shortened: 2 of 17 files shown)


alfonsograziano:
## Follow-up: making Codex work in the **sandbox** too (not just chat)

The PR ships Codex OAuth as **chat-only** because agentic-pi's `RunOptions` has no `apiKey`/`auth` field and Codex isn't in pi-ai's env-var auth map. I dug into whether that's actually a hard limit — it isn't.

### Finding: the capability already exists, it's just unplumbed

agentic-pi builds its agent on `@earendil-works/pi-coding-agent`, whose `AuthStorage` already does full OAuth — reads `auth.json`, refreshes tokens with file-locking, and resolves the Codex model's `chatgpt.com/backend-api` base URL through `ModelRegistry.create(authStorage)`. Two things line up:

- `AuthStorage` reads `<agent-dir>/auth.json`, and agent-dir is overridable via **`PI_CODING_AGENT_DIR`**.
- Its on-disk format is `{ "<provider>": { "type": "oauth", access, refresh, expires } }` — **byte-identical** to what this PR's `src/engine/oauth.ts` already writes to `$STATE_DIR/auth.json`.

### Proof (zero agentic-pi changes)

Pointing `PI_CODING_AGENT_DIR` at the Lastlight store and calling `agentic-pi run()` on the `none` sandbox with a Codex model:

```js
process.env.PI_CODING_AGENT_DIR = resolve("data"); // holds the openai-codex login
const result = await run({
  model: "openai-codex/gpt-5.5",
  prompt: "Reply with exactly the word: PONG",
  sandbox: "none", noBuiltinTools: true, maxSteps: 1,
});
// → finalText: "PONG"  | error: none
// → session,agent_start,message_end,agent_end,usage_snapshot
```

Clean full agent turn, authenticated via the subscription token, no code change.

### So the real backend matrix is:

| Backend | Codex in sandbox? | Why |
|---|---|---|
| `none`, `gondolin` | ✅ works today | model call is in-process; `AuthStorage` reads the host store via `PI_CODING_AGENT_DIR` |
| `docker` | ❌ not yet | model call runs *inside* the container, so `auth.json` must be staged into the guest |

The docker gap is the only thing needing an agentic-pi change, and it's small: add a `RunOptions.authFile` (or `oauthCredentials`) that agentic-pi materializes into the guest's `~/.pi/agent/auth.json` (or sets the guest `PI_CODING_AGENT_DIR` to a mounted path) — mirroring how it already auto-injects `GITHUB_TOKEN` into `sandboxEnv`. For `none`/in-process it's a no-op.

### Options to proceed

1. **Lastlight quick-win only** — set `PI_CODING_AGENT_DIR` in the sandbox env here so Codex works on `none`/`gondolin`, and narrow the "chat-only" warning to docker-only. Ships in this PR; covers local + native(gondolin). No agentic-pi change.
2. **Both repos (full fix)** — the quick-win **plus** an `agentic-pi` PR adding `RunOptions.authFile` that stages the store into the docker/gondolin-guest sandbox, so Codex works on every backend incl. production docker. Two PRs.
3. **Write-up only** — leave this PR chat-only for now and treat the above as a design note for a separate effort.

@cliftonc — since you own agentic-pi, which route do you prefer? Happy to open the agentic-pi PR if you want option 2.

cliftonc:
Best to also fix it on agentic-pi - it's a good addition, it's also on the Nearform org.

cliftonc:
Thanks @alfonsograziano 🚀

PR #165, opened 2026-07-13
feat(classifier): fork the classifier prompt + workflow-driven intents (#164)

### LINKED ISSUE #164: Fork the classifier prompt
Currently the classifier is a critical part of core lastlight, that decides how to route free text content from a user (chat or github) to the right workflow.  It is very important that this prompt is also possible to be forked as part of the agent-context.

I also want to ensure that the classifier works nicely as we add new workflows.  I wonder if we can also consider the idea that a workflow exports its own classification template?  And the classifier imports all of those?

cliftonc:
Closes #164.

## Problem
The intent classifier — which routes all free-text input (GitHub comments, Slack messages) to a workflow — was a single hardcoded ~95-line `CLASSIFIER_PROMPT` string in `src/engine/screen/classifier.ts`. It baked together framing, 14 category descriptions, and disambiguation rules, so it was **not forkable** per-deployment and **did not scale** as workflows were added (every new workflow meant hand-editing the one string).

## What this does
The classifier prompt is now **composed at runtime**, not hardcoded:

- **Forkable base template** — `workflows/prompts/classifier.md`, resolved through the existing overlay/`resolvePromptPath` machinery (overlay wins by name). Holds the framing, global disambiguation rules, the 5 harness-owned control categories (`APPROVE`/`REJECT`/`STATUS`/`RESET`/`CHAT`), and the `{{categories}}`/`{{examples}}`/`{{intentTokens}}` slots.
- **Per-workflow categories** — each of the 9 intent-owning workflows carries its own `classification:` block (`intent` + `description` + `examples`) in its YAML. `build.yaml` owns `BUILD`, `pr-review.yaml` owns `REVIEW`, etc.

### An overlay workflow "just works"
The token→intent parser vocabulary **and** an intent→workflow `getWorkflowByIntent` router fallback are derived from the same `classification:` blocks. So adding a workflow (even in an overlay) — e.g. `instance/workflows/incident.yaml` with `classification.intent: incident` — teaches the classifier the `INCIDENT` category, the parser the token, and the router the route, **with no core edit**.

The router's existing per-intent branches are untouched (they encode context-dependent logic like `build`→`pr-fix` on a PR); only the trailing generic default falls back to `getWorkflowByIntent`, and only for intents *outside* the well-known set — so `explore`-on-a-PR etc. keep their exact behavior.

**Limitation (documented):** a brand-new intent routes to its single owning workflow uniformly across GitHub issue/PR/Slack; surface-specific routing still needs `routes.*` overrides.

### Consolidation
- `ISSUE_QUESTION_PROMPT` retired — new issues route through the main classifier's `QUESTION` intent (`answer.yaml` owns it) via `classifyIssueIntent`.
- The re-triage gate is now the forkable `workflows/prompts/classify-adds-info.md`.
- `lastlight fork classifier` forks the base prompts (per-workflow category text already travels with `fork <workflow>`).
- Loader `validateAssets` rejects duplicate / reserved-control / token-colliding `classification.intent` at boot; composition is memoized against a loader asset-version counter (no cross-layer import).

## Verification
- Full suite: **1061 passed, 7 skipped**; `tsc --noEmit` clean.
- **Prompt parity** confirmed — the composed output is structurally byte-identical to the original 14-category prompt.
- **End-to-end `incident` acceptance check** run against a real overlay (no mocks): category composes in, token line includes `INCIDENT`, `getWorkflowByIntent` resolves, `classifyComment` parses it.
- New tests: classifier composition + `classifyIssueIntent`; loader validation + `getWorkflowByIntent`; router novel-intent fallback (GitHub + Slack, incl. unmanaged-repo rejection + a "well-known excluded" negative case); fork-cli `classifier` target.

## Docs
- This repo: `spec/05-router.md`, `spec/06-workflow-engine.md`, `CLAUDE.md`, `src/workflows/CLAUDE.md`.
- Separate `lastlight-www` repo: `docs/cli.astro` gets `lastlight fork classifier` (committed separately).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -148,8 +148,15 @@ src/
     screen/             Prompt screening + intent classification.
       screen.ts         Prompt-injection screener. Uses llm.ts with a cheap
                         model (claude-haiku by default).
-      classifier.ts     Tiny LLM call that decides "is this comment asking
-                        me to build something?". Uses llm.ts.
+      classifier.ts     Tiny LLM call that classifies a comment/message into
+                        an intent (build / review / … / chat). Uses llm.ts.
+                        The prompt is COMPOSED at runtime, not hardcoded: a
+                        forkable base (workflows/prompts/classifier.md) + one
+                        `classification:` block per workflow YAML. Adding a
+                        workflow (even in an overlay) adds a routable intent
+                        with no core edit — the router's getWorkflowByIntent
+                        fallback routes it. `lastlight fork classifier` forks
+                        the base prompt. (issue #164)
   workflows/            See src/workflows/CLAUDE.md for the full runner
                         story. Loads YAML definitions, executes phases
                         (linear or DAG), manages resume, approval gates,
@@ -502,6 +509,9 @@ lastlight server status                # compose ps + core/overlay version drift
 lastlight fork                         # list forkable workflows + agent-context (marks forked)
 lastlight fork <workflow>              # workflow YAML + every prompt + skill its phases reference
 lastlight fork agent-context [file]    # all agent-context/*.md (soul/rules/security), or one file
+lastlight fork classifier              # the base intent-classifier prompts (classifier.md +
+                                        # classify-adds-info.md); per-workflow category text
+                                        # travels with `fork <workflow>` already
                                         # [--home dir] [--force to overwrite existing]
 
 # Install the Last Light Claude Code skills into a local Claude Code (HOST-LOCAL;
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -50,9 +50,13 @@ Codex model is used for a sandbox phase).
 
 The cheap-helper path (`src/engine/llm.ts`, used by screener + classifier)
 bypasses OpenCode and dispatches directly to the same three providers.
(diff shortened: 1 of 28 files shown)


cliftonc:
## Added: classifier/screener model now settable via `config.yaml`

Follow-up to a review note — the cheap one-shot helpers (`classifier`, `screener`) resolved their model only through `defaultFastModel()` → `OPENCODE_MODELS` env + provider order, **ignoring** the `config.yaml` `models:` map that workflow phases use. So `models.classifier` in `config.yaml` was a silent no-op and the only way to pin the helper was an odd env var.

`defaultFastModel(taskType)` now resolves in order:
1. `config.models[taskType]` (`models.classifier` / `models.screener` in `config.yaml`; env `OPENCODE_MODELS`/`LASTLIGHT_MODELS` is already layered into this map, so env still wins);
2. the env `OPENCODE_MODELS` map directly (pre-config-load / CLI contexts);
3. the first configured provider's fast model.

Only an **explicit** per-task entry counts — never `models.default` — so the helpers stay cheap unless deliberately pinned. Docs updated (`spec/02`, `spec/05`, `CLAUDE.md`, commented example in `config/default.yaml`). +2 tests; full suite 1063 passing.

Issue #210, opened 2026-07-22
Missing toolchain when running Last Light in k8s


robinbowes:
# Gondolin (the default sandbox) doesn't run in the published Docker image / K8s: no QEMU, and the guest image lacks the security tooling the docker backend bakes in

## Summary

Deploying Last Light to Kubernetes with the **default** sandbox backend
(`gondolin`) surfaces two gaps. Neither is a misconfiguration — the published
`lastlight-agent` image and the gondolin guest image are missing pieces that the
`docker` backend gets from the sandbox images. The docs
([run-it](https://lastlight.dev/run-it/) step 5) state *"the same Docker image
works in K8s,"* but with the shipped default that isn't true.

Filing so the k8s/gondolin path can be made first-class (or the limitation
documented). I have a working deployment and am happy to contribute.

## Environment

- Kubernetes on Talos Linux (containerd), bare-metal amd64 nodes with `/dev/kvm`
- `ghcr.io/nearform/lastlight-agent:v0.21.7`
- `sandbox.backend: gondolin` (the `config/default.yaml` default)
- No Docker socket available (containerd), so the `docker` backend isn't an option

## Gap 1 — the published image has no QEMU, so the default backend can't run

Every workflow phase fails immediately:

```
[agentic] --sandbox=gondolin failed (qemu-not-installed): Install QEMU on the
host: 'brew install qemu' (macOS) or 'apt install qemu-system-x86
qemu-system-arm qemu-utils' (Debian/Ubuntu). See SPIKE-gondolin.md for context.
  [executor] Result: error_exit_2 (0 turns, 12s)
```

The `lastlight-agent` image ships no `qemu-system-x86_64`. Gondolin is the
*default* backend, but the image is built for the `docker` backend (it has the
docker group and expects `/var/run/docker.sock`); gondolin's QEMU is only
provided on the *native* systemd host (`deploy/native/install.sh` `apt install`s
it). So a user who follows the K8s docs with the default backend gets a
non-working sandbox.

### What made it work (offered as the basis for a fix / docs)

1. **QEMU in the image** — a derived image:
   ```dockerfile
   FROM ghcr.io/nearform/lastlight-agent:v0.21.7
   USER root
   RUN apt-get update \
    && apt-get install -y --no-install-recommends qemu-system-x86 qemu-utils \
    && rm -rf /var/lib/apt/lists/*
   ```
2. **`/dev/kvm` to an unprivileged pod** — via the Talos-recommended
[...]

PR #211, opened 2026-07-23
feat: publish lastlight-agent-qemu image for the gondolin/k8s path (#210 Gap 1)

robinbowes:
Fixes Gap 1 of #210. The published `lastlight-agent` image ships no QEMU, so the
default `gondolin` sandbox backend can't run in a container/Kubernetes
environment — every workflow phase fails with `qemu-not-installed`. The
`run-it` k8s docs ("the same Docker image works in K8s") don't hold for the
shipped default.

## What this does

- **New `lastlight-agent-qemu` image** (`apps/server/agent-qemu.Dockerfile`):
  base `agent` + `qemu-system-x86` + `qemu-utils`. Wired into `docker-bake.hcl`
  as a content-keyed context target (the same pattern `sandbox` uses against
  `sandbox-base`) and added to the `core` bake group, so `publish.yml` builds
  and pushes it on release with `:vX.Y.Z`/`:latest` — no workflow change.
- **k8s example** under `apps/server/deploy/k8s/`: a `kubectl apply -k`-able set
  (namespace at `privileged` PSA, the `squat/generic-device-plugin` DaemonSet
  advertising `devic.es/kvm`, Deployment/PVC/Service/ConfigMap) plus a README.
  The pod stays unprivileged apart from one `SYS_RESOURCE` capability + the KVM
  device — no `privileged: true`, hostPath, or host namespaces.
- **Docs**: `run-it` Step 5 rewritten to explain the backend/image split and
  point at the example; `RELEASING.md`, `apps/server/CLAUDE.md`, the spec, and
  the bake/publish comments updated to record the fifth image (noting it is
  published but not pulled by the compose `server update` flow).

## Design note for review

QEMU ships as a **separate `-qemu` variant** to keep the base image lean for
docker-backend users (QEMU adds a few hundred MB they never use), matching the
proposal in #210. Since gondolin is the *default* backend, there's a reasonable
alternative: bake QEMU into the base `lastlight-agent` so the default works
everywhere out of the box. Happy to switch to that if you'd prefer — it's a
small change (move the apt layer into `apps/server/Dockerfile`, drop the variant
target). Your call.

## Scope

Gap 2 (the gondolin guest image lacks `gitleaks`/`semgrep`, so `security-review`
degrades to a manual pass) is **not** in this PR — a follow-up. #210 stays open;
the README and `run-it` note the limitation.

## Verification

- `docker buildx bake --print core` lists the new `agent-qemu` target with the
  content-keyed `agent` context and correct tags.
- Built the image locally (amd64): `qemu-system-x86_64` present (QEMU 7.2.22).
- `kustomize build` + `kubectl apply -k --dry-run` validate the example set.
- Against a Talos/KVM cluster with this recipe, `repo-health`,
  `security-review`, and `issue-triage` complete in gondolin micro-VMs — the
  deployment this example generalizes from.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -30,8 +30,8 @@ name: Publish release
 # behind `checks`); it never publishes to npm — the `npm` job is gated to
 # `release` events.
 #
-# Build the four locally-built images (agent + sandbox-base + sandbox +
-# sandbox-qa) once and push them to GHCR, so `lastlight server update` PULLS them
+# Build the five locally-built images (agent + agent-qemu + sandbox-base +
+# sandbox + sandbox-qa) once and push them to GHCR, so `lastlight server update` PULLS them
 # on the deploy host instead of building from source (minutes → seconds).
 #
 # The stock sidecar images (coredns / nginx / otel-collector / caddy) are pulled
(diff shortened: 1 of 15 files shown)


cliftonc:
Looks great @robinbowes thanks!

Issue #213, opened 2026-07-23
pr-review cron job 404s on private repos


robinbowes:
## Summary

The scheduled dependency-PR crons (`cron-dependabot-ci-fix`,
`cron-dependabot-merge`) return `404` when enumerating open PRs on **private**
repos, while everything else works: webhook-driven `pr-review` reviews the same
private repo successfully, and the crons handle every public repo fine.

The cause is a scope mismatch between two code paths:

- The **webhook / agent** path mints a **per-repo scoped** installation token
  (`agent-executor.ts`: `repositories = access.repo ? [access.repo] : undefined`)
  and works on private repos.
- The **cron enumeration** path lists PRs server-side via
  `GitHubClient.listOpenPullRequests()` / `getRepository()`
  (`github.ts` → `this.octokit.rest.pulls.list` / `repos.get`) using an
  installation client that, in this path, does **not** cover the private repos.

**Public repos hide the bug.** An installation token can read/clone any public
repo regardless of whether it's in the token's scope, so the cron's `GET
/repos/{owner}/{repo}/pulls` succeeds on every public repo even with a narrow
token. Only private repos exercise the token's real scope — so the failure is
invisible on public repos and surfaces only on private ones.

Filing because the failure is silent (no error surfaced to the operator; the run
just records a skip) and asymmetric (private-only), which makes it hard to
notice. I have a working multi-repo deployment and am happy to contribute a fix.

## Environment

- Kubernetes (Talos Linux, containerd), bare-metal amd64
- `lastlight-agent` v0.21.7, `sandbox.backend: gondolin`
- GitHub App installed on an **organization** account, `repository_selection:
  "all"` (confirmed via `GET /app/installations/{id}`)
- `managedRepos`: 14 repos (12 public + 2 private)

## Symptom

Manually running the dependency-PR cron across all managed repos, from the
server log (public repos elided):

```
[executor] Minting git token: profile=review-write, repo=jobhound, permissions=contents,issues,pull_requests,metadata
[executor] Minting git token: profile=review-write, repo=gh-release-stats, permissions=contents,issues,pull_requests,metadata
GET /repos/OWNER/private-repo-a/pulls?state=open&per_page=30 - 404 with id … in 301ms
GET /repos/OWNER/private-repo-a                             - 404 with id … in 246ms
[executor] Minting git token: profile=review-write, repo=github-repos, permissions=contents,issues,pull_requests,metadata
[...]

PR #221, opened 2026-07-24
fix(pr-review): treat a scan with no PR to review as a clean no-op

robinbowes:
## Problem

A cron `mode: scan` pr-review (`cron-review.yaml` → `check-prs-awaiting-review`) over a repo with **no open PRs** fails the run at `post-review` with:

```
post-review: no PR number in run context; cannot post review
```

The review phase completes fine (nothing to review), but `PhaseExecutor.runPostReview` checks `prNumber` **before** reading the findings file and honouring `doc.skip`. In scan mode no PR is handed in, so a legitimate "nothing to review" can never reach the skip branch — the run red-Xes. Every quiet managed repo therefore fails every cron tick.

This is the actual cause behind reports that the pr-review cron "fails on private repos": a repo with no open PRs fails regardless of visibility; private repos are just the ones an operator notices (and, on a token-less code path, a private repo additionally 404s on the listing — tracked separately). The visible red X is this post-review false-failure.

## Fix

- **`post-review`**: read `findings.json` and honour an explicit `doc.skip` **before** the `prNumber` gate. A deliberate skip is a clean no-op even when scan handed in no PR. A **missing** findings file still fails loudly — so a review that couldn't actually run (e.g. the agent errored listing PRs) is surfaced, not silently swallowed.
- **`pr-review` skill**: instruct the agent to write `{skip, summary}` when a scan finds **no open PRs** — and explicitly **not** when listing *errors* — so the no-op is a deliberate skip rather than an absent file.
- **spec** (`06-workflow-engine.md`): document the skip-before-prNumber ordering.

## Tests

`tests/workflows/post-review.test.ts`:
- new: a scan with `{skip:true}` and no PR number **succeeds as a no-op** (nothing posted) — fails before this change with the exact `no PR number` error.
- unchanged: the four existing posting tests still pass (targeted-PR posting, missing-findings-after-a-real-review still fails, etc.).

Full `tests/workflows/` suite green (301 tests). No new type errors.

## Scope

Behavioural fix + skill contract. Related to the pr-review-cron discussion in #213 (the private-repo framing there resolves to this + a separate token-less-retry path, which is not in this PR).

### DIFF
--- a/apps/server/skills/pr-review/SKILL.md
+++ b/apps/server/skills/pr-review/SKILL.md
@@ -46,7 +46,11 @@ If `prNumber` (or `issueNumber`) is set in the Context block, **that is your
 target** — go straight to `github_get_pull_request` with it. Do **not** call
 `github_list_pull_requests` to "find" or "confirm" it; you were handed it, and
 listing dumps a large payload for nothing. Only when no PR is given (a repo-wide
-`mode: scan`) do you list open PRs and pick the most recent unreviewed one.
+`mode: scan`) do you list open PRs and pick the most recent unreviewed one. **If
+the scan finds no open PRs, write `{"skip": true, "summary": "no open PRs to
+review"}` and stop** — an empty repo is a clean no-op, not a failure. (Do this
+only when the list genuinely comes back empty; if listing *errors*, do NOT write
+a skip — let it fail so the real problem surfaces.)
 
 **Stop conditions** (check before reviewing):
 - PR authored by `last-light[bot]` → skip. Never self-review.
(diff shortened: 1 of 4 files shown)


robinbowes:
Superseded by #222, which repairs the pr-review cron by migrating it off `mode: scan` entirely (discovery + per-PR fan-out) rather than patching the scan path. On that path an empty repo simply produces no runs, so the no-op this PR adds is no longer needed.

Keeping this open as the fallback until #222 merges — if that stalls, this is still the standalone fix for the empty-repo `post-review` failure. Please close in favour of #222 once it lands.

PR #250, opened 2026-07-30
fix(executor): thread per-run GitHub credentials explicitly, not via process.env (#215)

### LINKED ISSUE #215: Agent GitHub *writes* from inside the sandbox 403, while identical harness-side writes succeed (dependabot crons can't act)
## Summary

The `dependabot-pr-merge` / `dependabot-ci-fix` flows fail on every repo: the
agent's GitHub **write** calls made *during the run* — create-label, merge,
comment — return `403 "Resource not accessible by integration"`. The same
operations, with the same App and a correctly-minted `repo-write` token,
**succeed** when issued outside that path (a hand-minted copy of the token, and
the harness-side `pr-review` `post-review` comment).

The difference is *where the write originates*:

| Flow | Where the write happens | Result |
|---|---|---|
| `pr-review` | **harness-side** `post-review` (`GitHubClient.withToken(review-write)`) | ✅ comment posts |
| `dependabot-pr-merge` / `-ci-fix` | **agent-side, during the run** (agentic-pi GitHub tools) | ❌ 403 on label / merge / comment |

App permissions, token scope, and the mint itself are all provably correct
(evidence below), so this is not a permissions/config problem — it's specific to
agent writes originating inside the sandbox.

## Environment

- Kubernetes (Talos), `sandbox.backend: gondolin`, `lastlight-agent` v0.21.7
- GitHub App on an **org**, `repository_selection: "all"`
- App repository permissions: Contents R/W, Issues R/W, Pull Requests R/W,
  Workflows R/W, Checks R/W, Metadata R (all granted and org-accepted)

## Symptom (server log)

```
[executor] Minting git token: profile=repo-write, repo=REPO,
           permissions=contents,issues,pull_requests,workflows,metadata
  [executor] Running agent (task: REPO-N-dependabot-pr-merge, sandbox: gondolin)
POST /repos/OWNER/REPO/labels              - 403   ← create label definition (repo-level)
POST /repos/OWNER/REPO/issues/N/labels     - 403   ← add label to PR
PUT  /repos/OWNER/REPO/pulls/N/merge       - 403
POST /repos/OWNER/REPO/issues/N/comments   - 403
```

Every 403 body is `{"message":"Resource not accessible by integration", …}`.

## Why it isn't permissions / scope / staleness (all ruled out)

1. **The mint is correct.** The log line above shows the token minted with
   `contents,issues,pull_requests,workflows,metadata` write/read.
2. **A copy of that exact token works.** Hand-minting the same `repo-write`
   scoped token and calling `POST /repos/OWNER/REPO/labels` returns **`201`**;
   `X-Accepted-GitHub-Permissions: issues=write; pull_requests=write` confirms the
   token's permissions satisfy the endpoint.
3. **Harness-side writes with a *narrower* token work.** `pr-review`'s
   `post-review` posts its comment via `GitHubClient.withToken(<review-write
   token>)` (`contents:read, issues:write, pull_requests:write, metadata:read`)
   and succeeds — on a live PR, same repo, same run window in which the dependabot
   writes 403.
4. **Not stale auth.** `kubectl rollout restart` then re-run — identical 403s.
5. **Not a Dependabot-PR restriction.** One of the failing calls,
   `POST /repos/OWNER/REPO/labels`, creates a **repo-level** label definition — it
   is not scoped to the Dependabot PR — and still 403s. So it isn't GitHub
   restricting apps on Dependabot-authored PRs; it's the token/path.

## What the source shows

The agent's GitHub client (`packages/agentic-pi/src/extensions/github/`) obtains
its token via `buildAuthFromEnv`:

- `GitHubAppAuth.getToken()` mints an installation token with **no body** → a
  full-installation token (**all** App permissions). That would *succeed* at
  every failing call.
- `StaticTokenAuth.getToken()` returns `GITHUB_TOKEN` — the scoped `repo-write`
  token. That *also* has `issues/pull_requests/contents:write` and succeeds when
  used directly (evidence #2).

**Both** possible in-sandbox credentials should work, yet the calls 403 with an
*integration* token — so the credential actually presented at the agent's call
site is neither a full-install token nor the intact scoped `repo-write` token. It
appears to be an App installation token that has been narrowed to read somewhere
between the executor's mint and the outbound request (the gondolin egress path /
how the agent's tool calls are executed). That layer is where the bug is.

## Reproduce

1. Managed repo with an open Dependabot PR; App with all the perms above.
2. Run `merge-green-dependency-prs` (or let `pr.checks_passed` fire).
3. During the run, the agent's `POST …/labels`, `PUT …/pulls/N/merge`, and
   `POST …/issues/N/comments` all 403.
4. Contrast: a `pr-review` on any PR posts its comment fine (harness-side write).

## What would pin it

Log, at the agent's outbound GitHub request, the **token id / permission set
actually presented** (or the response's `X-Accepted-GitHub-Permissions` alongside
the token's granted permissions). That distinguishes "the scoped token is being
narrowed in transit" from "a different, read-scoped token is selected for the
in-sandbox client."

## Impact

Bounded: only the `dependabot-pr-merge` / `dependabot-ci-fix` flows are affected
(they can't label/merge/comment on any repo). Webhook `pr-review`, `issue-triage`,
and `repo-health` all work, on public and private repos.

Happy to help test a fix or add request-level logging to pin the credential.

cliftonc:
Fixes #215.

cc @robinbowes — this is the root cause behind the 403s you reported, and it explains why your clean retest passed: that run had a single green PR, and a single non-overlapping run never trips this. Your evidence was decisive in locating it — a correctly write-scoped mint in the log, a hand-copied token that worked, and harness-side writes succeeding in the same window meant the credential presented at the agent's call site was neither the intact scoped token nor a full-install one. It turned out not to be the gondolin egress path but the harness, one layer up.

## The bug

The token really was being narrowed, in `InProcessSandbox.runAgent`. For the in-process backends (`gondolin`/`none`) the agent runs **in the harness process**, so the run's credentials were spliced into the **global** `process.env` and restored afterwards. With `concurrency.maxWorkflows: 4`, up to four runs are live in that one env, and the splice is not re-entrant. Two failure modes followed:

**1. A run captured another run's token.** agentic-pi reads the env *late* — `loadGitHubExtension` → `buildAuthFromEnv(process.env)` runs only after `ModelRuntime.create()` + `modelRegistry.refresh()` (real file IO). A second run splicing inside that window left the first holding the *other* run's token for its whole session: wrong repo, and read-only if that run's profile was `read`/`issues-write`/`review-write`. Every `github_*` write then 403'd, while `git clone`/`push` kept working — those get the token explicitly via `sandboxEnv`, not from `process.env`. That asymmetry is exactly the reported symptom.

**2. Interleaved restores permanently poisoned the harness env.** Run B saved the values run A had spliced, so B's restore reinstated *A's* rather than the originals:

```
boot                         {"APP_ID":"12345","PEM":"/app/data/secrets/app.pem"}
A spliced                    {"APP_ID":"","PEM":"","TOKEN":"ghs_A_write_repoX"}
B spliced (A sees B token)   {"APP_ID":"","PEM":"","TOKEN":"ghs_B_read_repoY"}
A restored (B mid-run)       {"APP_ID":"12345","PEM":"/app/data/secrets/app.pem"}
B restored -> TERMINAL       {"APP_ID":"","PEM":"","TOKEN":"ghs_A_write_repoX"}
```

This never healed: the next splice saved the poisoned values and restored them again. And `prepareRun` gated minting on `process.env.GITHUB_APP_ID`, now falsy — so it **skipped the mint entirely** and took the PAT fallback, forwarding that stale leftover token to the agent *and* the pre-clone. From then on every in-process run reused one dead token: wrong repo, wrong profile, expired within the hour.

Why the dependabot flows took the hit: they're the ones whose writes happen **agent-side, during the run**. `pr-review`'s comment is posted harness-side by `post-review`, which never reads the live `process.env` — the same asymmetry as the hand-copied token working. And `POST /repos/{owner}/{repo}/labels` 403'ing was never Dependabot-specific: a read-scoped or wrong-repo token 403s there regardless of target.

This is the same race `post-review` was already patched around one layer up (the `PEM=""` → `resolve("")` → EISDIR crash).

## The fix

**1. Per-run credentials never travel through `process.env`.** agentic-pi gains `RunOptions.githubAuthEnv` — the run's GitHub credentials, **replacing** `process.env` for the `github_*` extension rather than merging with it. So an empty value means "this run has no credential", not "fall back to the ambient env" — which is how the harness's App PEM or a host PAT could otherwise reach an agent whose profile was deliberately downscoped. `InProcessSandbox.runAgent` now splices `withoutGitHubCredentials(env)` and passes `githubAuthEnv: githubAuthEnvFrom(env)`. The container backends (docker/k8s/smol) always had per-run envs and are untouched; the `Sandbox` port is unchanged.

**2. Minting is decided from boot config**, `getRuntimeConfig().githubApp` via `resolveGithubApp()`, never live `process.env` — mirroring the `post-review` fix — so the silent stale-token fallback can't return even if something re-introduces an env mutation. It also refuses to mint when `githubApiBaseUrl` is set, keeping the evals' mocked GitHub off any real installation now that the decision is no longer env-driven.

**3. The PAT fallback is boot-config-first** and its warning now states plainly that the token was *not* minted for this run. I deliberately did **not** make it hard-refuse `repo-write`: PAT-only installs use that on purpose, and the evals run code-fix tiers on a static fake token, so refusing would break both. The silent-degradation path it was guarding is closed structurally instead — the mint can no longer be skipped by a raced env var.

Plus `task=<taskId>` in the mint log, since being unable to attribute interleaved mints to runs is what made this hard to pin (it's also the "no mint line before the 403s" signature to look for in an affected install).

## Verification

`apps/server/tests/engine/agent-executor.concurrent-github-creds.test.ts` overlaps two runs (different repos, `repo-write` vs `read`) with distinguishable minted tokens and asserts each agent received its own credential, that `process.env` is untouched mid-flight and after, and that only the scoped token — never the App key — reaches the agent. Confirmed it fails against the pre-fix code with the exact mechanism:

```
× hands each overlapping in-process run its own minted token
    AssertionError: expected undefined to deeply equal { GITHUB_TOKEN: 'ghs_repo-a' }
× never mutates the shared process.env — no token splice, no App-key clear
    AssertionError: expected 'ghs_repo-b' to be undefined   ← run B's token, in run A's env
```

`packages/agentic-pi/test/extensions/github/auth-env.test.ts` covers the replace-not-merge semantics. Full workspace gate green (`pnpm turbo run typecheck test build`, 21 tasks).

## Docs

`spec/09-sandbox.md` gains an invariant section (which credential moves on which per-run channel, plus the post-mortem) and drops the now-false claim that low-trust sandboxes get `GITHUB_APP_PRIVATE_KEY_PATH=""`. agentic-pi's README previously recommended the broken pattern outright — "if your orchestrator runs multiple workflows with different credentials, `process.env` is the seam to vary" — now replaced with `githubAuthEnv`. `apps/server/CLAUDE.md` and the stale `post-review.ts` comment updated too.

## Rollout

This is prod-facing, so it needs a release to reach an instance (`server update` pulls the GHCR images a Release builds). @robinbowes: the pre-fix damage on an affected install is runtime-only — restarting the agent clears the poisoned env — and after this deploy there's no such state to clear, so `concurrency.maxWorkflows` no longer needs holding at 1 as a workaround.

### DIFF
--- a/apps/server/CLAUDE.md
+++ b/apps/server/CLAUDE.md
@@ -339,12 +339,19 @@ dashboard/              React+Vite admin SPA, served from /admin at runtime.
 - **Permission profiles** (`src/engine/github/profiles.ts`) — each workflow maps to
   a `GitAccessProfile`: `read`, `issues-write`, `review-write`, `repo-write`.
   `runner.ts` picks one per workflow name and the agent-executor mints a
-  downscoped installation token for the sandbox. Only `repo-write` runs see
-  the App PEM; everything else uses a pre-minted scoped token, which
+  downscoped installation token for the sandbox (minting is gated on **boot
+  config**, `getRuntimeConfig().githubApp`, never live `process.env`). No profile
+  forwards the App PEM today; every run uses that pre-minted scoped token, which
   agentic-pi's built-in github tools (its `github` extension — the
-  `github_*` tools, gated per profile) read from the sandbox env. The
-  standalone `mcp-github-app` MCP server that used to expose these tools was
-  removed with the OpenCode→agentic-pi migration.
+  `github_*` tools, gated per profile) read from a **per-run** credential
+  channel: the container backends get it in the container env, and the
+  in-process backends (gondolin/none) get it via agentic-pi's `githubAuthEnv`
+  argument. It is **never** spliced into the harness's shared `process.env` —
+  concurrent in-process runs live in that one env, so a token there crossed
+  between runs and 403'd every `github_*` write (issue #215; see
+  `spec/09-sandbox.md` → "Invariant: per-run credentials never travel through
+  `process.env`"). The standalone `mcp-github-app` MCP server that used to
+  expose these tools was removed with the OpenCode→agentic-pi migration.
 - **Approval gates** — phases can declare `approval_gate: post_architect`.
   When hit, the run persists with `status: paused`, a row in
   `workflow_approvals`, and the user can resolve it via GitHub comment
--- a/apps/server/CLAUDE.md
+++ b/apps/server/CLAUDE.md
@@ -128,7 +128,7 @@ src/
                         executeDocker/executeSmol/executeInProcess twins).
       shared.ts         Backend-agnostic building blocks (RunResultAccumulator,
                         skill-bundle staging, server-artifact stage/harvest,
-                        finalizeFromRunResult, env splice).
+                        finalizeFromRunResult, githubAuthEnvFrom).
     dispatcher.ts       Routes classified events to workflow or chat handler.
     event-shim.ts       Translates agentic-pi events → Claude-SDK envelope jsonl.
     llm.ts              One-shot LLM helper for screen/ + classifier —
--- a/apps/server/spec/09-sandbox.md
+++ b/apps/server/spec/09-sandbox.md
@@ -532,6 +532,7 @@ result = await agenticRun({
   profile,                  // GitHub access profile — see below
   sandbox: backend === "gondolin" ? "gondolin" : "none",
   sandboxEnv,               // env forwarded into the agent's bash
+  githubAuthEnv,            // THIS run's GitHub credential (see below) — never process.env
   cwd: agentCwd,            // the pre-cloned repo (workspace root if not pre-cloned)
   noSession: true,
   skillPaths,               // per-phase skill bundle dirs, absolute (see Skills §)
@@ -647,7 +648,9 @@ Per phase:
 
 1. `refreshGitAuth()` (`git-auth.ts`) mints a GitHub App installation
    token downscoped to the profile's permissions. Optionally scoped to
-   a specific repository allowlist.
+   a specific repository allowlist. **Whether to mint at all is decided from
+   boot config** (`getRuntimeConfig().githubApp`, via `resolveGithubApp`), never
+   from live `process.env` — see the invariant below.
 2. The token (not the PEM) is forwarded into the sandbox via
    `GIT_TOKEN` and `GITHUB_TOKEN` env vars. Git operations authenticate
    with it through a **github.com-scoped `http.extraheader`** (Basic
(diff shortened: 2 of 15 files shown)


cliftonc:
Pushed 7cd749e — the remaining `applyEnv` splice is now gone entirely, not just narrowed. Worth recording why, since "agentic-pi reads provider keys from `process.env`" was a smell I'd preserved rather than fixed.

Once the GitHub credentials were excluded from it, what the splice still carried was provider API keys, web-search keys and OTEL settings — and `prepareRun` builds every one of those by copying **verbatim out of** `process.env` (`getOtelEnvForSandbox` is a filtered copy too). So on the in-process path it was writing back values that were already identical: a no-op with a race attached. Worse, it made that env *look* per-run scoped when it structurally cannot be, which is precisely the trap #215 fell into — the next person to add a per-run value there would have inherited the same crossover.

`InProcessSandbox.runAgent` now has one rule: the agent runs in the harness process, so `process.env` is read-only and everything per-run is an explicit `agenticRun()` argument — `githubAuthEnv`, `authFile`, `sandboxEnv`, `cwd`, `allowedHttpHosts`. `applyEnv` and `withoutGitHubCredentials` are deleted as unused. Pinned by a new case asserting `process.env` comes back byte-identical from an in-process run.

The distinction the spec now draws: a credential **scoped to one run** must be an argument; **process-wide configuration** is ambient by nature. Provider keys are the latter — one set per harness, identical for every run, which is exactly why #215 bit the GitHub token and never `ANTHROPIC_API_KEY`.

One honest limitation: this does **not** make model credentials explicit end-to-end. pi-ai resolves provider keys from the ambient env inside `ModelRuntime`, and `CreateModelRuntimeOptions` has no env-injection point, so agentic-pi can't seal that path — it could only add a preferred one on top (`ModelRuntime.setRuntimeApiKey`, which is per-runtime and therefore already per-run, since agentic-pi builds a fresh runtime per `run()`). Worth doing if per-run model credentials ever become a real requirement — BYO-key/multi-tenant deployments — but it fixes no bug today, so I've left it out.

PR #254, opened 2026-07-31
feat(config): per-repository configuration via a committed .lastlight/

### LINKED ISSUE #180: Per repository configuration
All of the current configuration of lastlight is global; so changing something (e.g. a workflow prompt, or a model config) applies to all.

It would be ideal if we can do two things:

1 - a yaml file in the root (similar to dependabot) to configure the specific repo - this would allow us to define a different model for a workflow step for example? Or disable a specific cron job or workflow for this repo?
2 - a per repo overlay folder (same as the core) that allows us to over ride prompts or skills (not workflows) for any workflow that operates on a checked out version of the repo (e.g. build or pr-fix) - so this would work via the same CLI fork process, just not into an overlay repo but in an actual code repo.

cliftonc:
Closes #180.

All configuration was global — changing a model or a prompt changed it for every managed repo. A repo may now commit a `.lastlight/` directory that overrides a bounded subset of config for itself alone.

```
.lastlight/
├── lastlight.yml            # models, variants, disabled.workflows, approval (add-only), crons
├── workflows/prompts/*.md   # prompts only — a repo may never contribute workflow YAML
├── skills/<name>/SKILL.md
└── agent-context/*.md       # additive only
```

## The security model

`.lastlight/` is **always read from the repo's default branch**, never a PR head and never the sandbox checkout. A PR must not be able to reconfigure the agent reviewing it. Three further guardrails follow from that:

- **Agent context is additive only.** A repo file whose basename collides with an operator-owned one (`soul.md`, `rules.md`, `security.md`, or anything the overlay added) is dropped with a warning — a repo cannot neuter the operator's security rules by committing an empty file with the right name.
- **`approval` is add-only.** A repo may add a gate, never remove one the operator set.
- **Operators bound the whole surface** via a new `repoConfig:` block (`enabled`, `allowKeys`, `allowedModels`, `allowAssets`). Precedence is `default → overlay → env → repo`.

Fetching goes through the App-authenticated GitHub client, so **private repos work**; an anonymous contents fetch would have 404'd silently and looked like "no config". Cached under `<stateDir>/repo-config/` with a ~60s TTL plus ETag/treeSha conditional requests, so a cron fanning out over 20 repos costs 20 conditional requests rather than 20 downloads.

**Failure mode: warn, drop the offending keys, run anyway.** A repo's config file never fails a run. Every rejection is structured, surfaced on the run, and visible in the dashboard.

## Concurrency

Asset resolution moves to a per-run `AssetResolver` (`createAssetResolver`) rather than mutating the loader's process-global layer stack — with `concurrency.maxWorkflows: 4` and a cron fan-out firing across every repo at once, mutating the global would race. The existing module-level exports are kept as a facade over a default instance, so there are **zero call-site changes**. A test runs two workflows concurrently against different repos with different model *and* prompt overrides, gated by a barrier so both are provably in flight.

## Crons

Crons gain a first-class `crons: { enable, disable }` block at **every** layer (default, overlay, repo). Legacy `disabled.crons` still works and is unioned in, so existing deployments are unaffected.

A globally-off cron now **keeps its tick registered** so a repo can opt back in, with participation resolved at tick time — that avoids any dynamic croner re-registration when a repo edits its config. The operator's un-overridable kill switch is dropping `crons` from `repoConfig.allowKeys`.

## Surfaces

- **CLI** — a new `lastlight repo` tree: `repo fork` (prompts + skills only, never workflow YAML), `repo config validate` (offline), `repo config show <owner/repo>`. This required moving the pure schema/validation/merge into `packages/shared`, since the CLI may never gain an edge to `core` — the dep-cruiser gate is the proof.
- **Dashboard** — a `config` sub-tab on the repo page showing the effective config with per-leaf provenance; `repo`-sourced leaves are the only solid badge, so "this came from the repo" is scannable.
- **Evals** — a `repo-config` tier and a fixture seam on the fake GitHub. Verified empirically that the repo's prompt is what reaches the model and its agent-context lands in `AGENTS.md`; the model's *compliance* is ungraded for want of a provider key in CI.

## Pre-existing bugs fixed in passing

None of these are caused by this change; all were found while working on it.

- Dependabot cron discovery bypassed per-repo participation entirely.
- The dashboard cron toggle unregistered the tick, so an opt-in silently did nothing until a restart.
- Three admin routes rebuilt cron contexts without `_cronName`, ignoring opt-outs.
- `redactPublic`/`SENSITIVE_KEY_RE` and the config layer merge were duplicated — a drifted copy of the first is a secret leak.
- `spec/02-configuration.md` drift: listed a removed `workflowDir`, omitted `managedRepos`/`routes`/`disabled`/`cleanup`/`kubernetes`/`githubToken`, wrong `SlackConfig`, a fabricated env default, stale line-number citations.
- `spec/03-integrations.md` claimed cron fan-out has a concurrency limit of 3 (it fires all at once; the admission controller is the only throttle).
- A flatly wrong public FAQ answer telling users to "add conditional logic to the skill files" to manage more repos.

## Verification

- `pnpm turbo run typecheck test build` — **21/21 tasks**, dep-cruiser clean, both boundary invariants intact (no `shared→core`, no `cli→core`).
- `lastlight-core` **1743 passed / 15 skipped**; CLI **102**; evals **44**; `astro check` 0 errors.
- **No pre-existing test was edited to make anything pass** — the two touched test files are additions only.

## Not done / known limits

- The eval proves the mechanism reaches the agent but does not grade model compliance — that needs a run with a real provider key.
- Repo layers are picked up on the next run (~60s TTL), not instantly; there is no `push` webhook subscription and adding one would need every installation to re-consent.
- Unrelated latent bug spotted and deliberately left alone: `config/default.yaml`'s routes include `verify`, `qa_test` and `demo`, but the hardcoded `defaultRouteConfig()` fallback omits all three.

Per the release policy this is prod-facing and changes the CLI, so shipping it needs a GitHub Release to build the GHCR images.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -62,7 +62,12 @@ workspace deps) consumed by `lastlight-core` + `lastlight-evals` (and the privat
 dashboard) via `workspace:*`. Invariants: **no edge from
 `shared`/`workflow-engine` back to `core`** (dep-cruiser gate, runs in
 `typecheck`); **the cli never gains an edge to `core`**. Turbo `^build` orders
-builds; there are no TS project references.
+builds; there are no TS project references. Those invariants are why some logic
+lives in `shared` rather than where you'd first look — e.g. the per-repository
+`.lastlight/` config schema + bounds + merge
+(`packages/shared/src/repo-config-schema.ts`, issue #180): core needs it at
+runtime and the CLI needs it offline for `lastlight repo config validate`, and
+`shared` is the only package both can reach.
 
 ## Commands (from the repo root)
 
--- a/apps/evals/CLAUDE.md
+++ b/apps/evals/CLAUDE.md
@@ -152,7 +152,8 @@ The release commit is conventionally just the two version-file lines
 | `src/discovery.ts` | Multi-root tier discovery (`tier.json` → `defaultWorkflow`). |
 | `src/init.ts` | `init` — scaffold + `gh repo create` an overlay+evals repo. |
 | `src/add-case.ts` | `add-case` — author an instance from a real GitHub PR/issue (`gh`+`git`: base/head SHAs, `test_patch`, red→green verdicts). |
-| `src/fake-github.ts` | In-process fake GitHub REST API (seeds fixtures, records mutations). |
+| `src/fake-github.ts` | In-process fake GitHub REST API (seeds fixtures, records mutations) **plus** the non-REST `fetchRepoConfigTree` seam for a repo's `.lastlight/`. |
+| `src/repo-config.ts` | Per-repo config layer (#180): reads a case's fixture tree and drives core's OWN resolver over it. The one file with deep `lastlight-core/dist/...` imports — see below. |
 | `src/seed.ts` / `src/grade.ts` / `src/metrics.ts` | Workspace seeding (vendored fixture, git-source `base_commit` checkout, OR pr-review PR-head checkout — all from the `./.eval-cache/` mirror) / grading (execution TAP, behavioral, + `gradeReview` judge) / token-cost roll-up. |
 | `src/judge.ts` | One-shot LLM client for `gradeReview` (pr-review only) — direct provider `fetch`, temp 0. `EVAL_JUDGE_MODEL` overrides `defaultJudgeModel()`. |
 | `scripts/import-martian.ts` | Import Martian's Code Review Bench offline set (50 PRs) into the `pr-review` tier (`gh`+`git`: resolves base/head, pins SHAs). |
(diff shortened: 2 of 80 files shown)


cliftonc:
Both findings were real — fixed in a9fc363.

**Control keys (`jobs.ts`, `cronContext`, "Run now").** Confirmed: `...def.context` was spread after the control keys at all three sites, so a cron YAML declaring `_cronName` or `_cronGloballyEnabled` won. Control keys are now injected last.

Two things I did differently from the suggested diff:

- **`context.repos` stays overridable.** That ordering is deliberate — a cron YAML is meant to be able to pin its own repo list — so `...def.context` still sits ahead of `repos` and only the two control keys moved. There's now a test pinning that, so it can't be "tidied" away later.
- **"Run now" strips rather than reorders.** *Absence* of `_cronGloballyEnabled` is the signal that lets the button fire a globally-disabled cron, so re-injecting it after the spread would have broken the button. The YAML's copy is destructured out instead.

**Byte cap.** Confirmed. Took the re-verify-after-download option: `content.length` is re-checked before admitting the file, and the true length is recorded as `size` rather than the tree's claim (which may be absent or wrong, and downstream cap checks key off it). Kept `continue` rather than `break` so a later small file is still admissible. The cheap pre-check still short-circuits before the network round-trip when the tree *does* report a size — the post-check is a backstop, not a replacement.

Both fixes are pinned by tests that were verified to fail without them (3/5 and 6/7 respectively; the ones that pass either way are the normal-path and deliberate-behaviour guards). Full gate green: 21/21 turbo tasks, 1755 core tests passed / 15 skipped.

Nice catches — particularly spotting the two further instances of the ordering bug in `routes.ts`, which I'd have fixed in one place and missed.

PR #258, opened 2026-08-01
feat: structured JSON logging with explicit levels for operational output

robinbowes:
## Summary

Operational logging in `lastlight-core` is currently ~320 raw `console.*` calls emitting unstructured text, which log pipelines (Vector→Loki, etc.) can't reliably assign a severity to. This replaces them with a small leveled logger that emits **structured JSON lines to stderr** with an explicit string `level`, so severity is a field to filter on rather than a substring to guess at.

## What changed

- **`LoggerPort`** — a dependency-free interface in `packages/workflow-engine` (the deepest dependency leaf), injected via its existing `EnginePorts`.
- **pino adapter** confined to `apps/server` only (`src/logging/logger.ts`). Because the port lives in the engine and the concrete logger stays in core, **`pino` never enters the published `lastlight` CLI's dependency tree** — enforced by the existing dep-cruiser boundary (engine may import only `zod`).
- **~320 `console.*` sites** in `apps/server` + the 9 in `packages/workflow-engine` converted to `logger("component").<level>(...)`, with a triaged level policy (genuine per-turn/hot-loop tracing demoted to `debug`; once-per-trigger stays `info`).
- **Contract:** one JSON object per line to **stderr** — `level` (string), `component`, `msg`, `err` (`{message,stack}`), `time` (ISO), plus `trace_id`/`span_id` stamped from the active OpenTelemetry span (Loki↔Tempo correlation) when one is active.
- **Config is env-direct:** `LOG_LEVEL` (default `info`) and `LOG_FORMAT` (`json`|`pretty`, default auto — `pretty` on a TTY, `json` otherwise, so containers get JSON automatically). `LOG_LEVEL` wired into the k8s deployment env; documented in `apps/server/CLAUDE.md`.

## Out of scope (deliberately)

- `packages/cli` and `apps/evals` are `chalk`/spinner **terminal UI**, not operator logs — left as-is.
- `agentic-pi` already routes diagnostics through an injected `onWarn`/sink and never touches `console`; only the harness-side consumer seam is repointed.
- **stdout is now reserved** for the sandbox NDJSON protocol; operational logs go to stderr exclusively.

## Behavior change for operators

Logs are now JSON on stderr (previously mixed plaintext across stdout/stderr). Plaintext `grep` workflows should move to a JSON-aware view or set `LOG_FORMAT=pretty` for local/human reading.

## Validation

- Full suite green (1762 tests); the pino adapter is TDD'd, including the OTel trace-mixin and the invalid-span-context / unrecognized-level edge cases.
- **Verified live in a Kubernetes homelab**: pods emit JSON with correct string levels; a Vector `pod_level` transform now reads `detected_level` from the body instead of guessing (clean before/after in Grafana), and workflow logs carry `trace_id` for Loki↔Tempo correlation.

## Note for reviewers

The branch currently includes design/plan docs under `docs/superpowers/`, which doesn't match this repo's `docs/*-design.md` + `docs/plans/` convention. Happy to relocate the design doc to `docs/structured-logging-design.md` and drop the implementation-plan artifact before review — flagging rather than assuming.


robinbowes:
@bot review

cliftonc:
Nice!  Really glad to have someone who knows how to operate things making improvements 😄

robinbowes:
> Nice! Really glad to have someone who knows how to operate things making improvements 😄

One tries one's best... 😄 

It would be good to get this merged ASAP so I don't need to keep rebasing it on all your releases. Any chance?

I'm currently rebasing on v0.23.7

cliftonc:
Will do it now!

robinbowes:
> Will do it now!

Excellent.

I'm currently rebasing all three PRs - #258, #260, and #263. Shouldn't be long.

Is there a good place in the repo to leave a note about using `logger` and not writing directly to console?

cliftonc:
Add it to the root Claude.md

robinbowes:
All done. This, #260, and #263 all ready for final review/merge.

cliftonc:
Ok will merge all three before moving the goal posts again!

robinbowes:
> Ok will merge all three before moving the goal posts again!

Thanks!

PR #274, opened 2026-08-06
fix(chat): keep Slack thread context across workflow-answered turns, and stop questions burning a sandbox

cliftonc:
Two related fixes to how a Slack thread is handled — one restores the thread's memory, the other stops most questions provisioning a sandbox to begin with.

## 1. The whole thread is recorded, not just its chat turns

Reported symptom: a question in a Slack thread was answered by the `answer` workflow; the follow-up "Can you summarise this?" in the *same thread* got

> There's no prior context in our conversation for me to summarise — this appears to be the start of our session.

A Slack thread is one conversation however each message was handled, but only `ChatRunner` ever wrote to `messaging_messages`. A message the classifier routed to a **workflow** left the history untouched, so the next chat turn rehydrated nothing. Confirmed on prod: that thread's chat session had `conversation_message_count: 1`.

Persistence moved into `ChatRunner` in 0d470fc to stop the chat path double-writing every turn — correct for chat, but `ChatRunner` runs on the chat path only. This restores the record for exactly the paths it doesn't own, with the two writers **mutually exclusive per turn**:

| Turn | Writer |
|---|---|
| `handler: chat` | `ChatRunner` (unchanged) |
| Any other messaging path — workflow dispatch, router refusal, approval/status reply | `withThreadTranscript`, wired once in `dispatch()` |
| A workflow's own output into the thread | `recordThreadMessageForThread`, from the runner's Slack `postComment` and the boot-recovery poster |

The wrap sits at the dispatch choke point rather than in per-handler code, so it covers **every** Slack-answering workflow — `answer`, `build`, `explore`, `verify`, `qa-test`, and any overlay-introduced intent.

### Two further bugs found on the way

- **Session staleness.** `touchSession` was likewise chat-only, so a thread carried by workflow turns lapsed past `SESSION_TIMEOUT_MS` and silently re-keyed to a fresh session mid-conversation. Every transcript write now touches; the by-thread writer also looks past the cutoff (`includeStale`), since a workflow can outrun the window between a question and its answer — recording revives the session so the thread continues.
- **`getHistory` returned the oldest 50, not the newest.** `ORDER BY timestamp ASC LIMIT 50` meant a long thread permanently rehydrated its own opening and never saw what was just said. Now newest-N reversed for the caller, with `id` as the tiebreak (a turn's two rows share a whole-millisecond timestamp).

Entries are clamped to 4000 chars keeping the **tail**, so a health report can't dominate the next turn's prompt.

## 2. `QUESTION` is now a capability test, not a seriousness one

`answer` provisions a sandbox; chat replies in-process in seconds. The old category asked whether a question was "substantive" and "benefits from research" — true of most technical questions, so anything naming a repo burned a container.

Chat is not toolless: it reads repos, issues and their comments, PRs **and their diffs**, file contents, commit history, and searches both code and issues. The category now admits only what chat genuinely cannot do — **the web** (comparisons with other tools, upstream docs, current external facts) and **real exploration of a checkout** (tracing behaviour across many files).

The old examples taught the opposite of the new rule: `"what's the difference between a sandbox phase and a chat session in cliftonc/lastlight?"` was a QUESTION example, and that's one `github_search_code` call. It's replaced by a trace-the-code-path example, plus three repo-scoped **CHAT counter-examples** — without a negative case the model reads `REPO: <x>` as the signal to fire.

**One carve-out, and it's load-bearing.** `classifyIssueIntent` reuses this same prompt for `issue.opened`, where `CHAT` doesn't mean "chat answers it" — it means **triage**. A blanket downgrade would have started filing question-issues as work items. So the downgrade is scoped to chat/Slack messages, taught by a deliberately near-identical `ISSUE TITLE` example that stays `QUESTION`.

## Testing

- Full core suite green: **2593 passed**, 20 skipped. `astro check`: 0 errors.
- New `thread-transcript.test.ts` (9 cases) over a real in-memory `SessionManager`, including that a reply failure still propagates and that a run outliving its session still records.
- 5 new `session-manager` cases (newest-N ordering, same-millisecond tiebreak, stale-vs-writer lookup).
- 4 new `dispatcher` cases — notably that the chat path is **not** double-recorded and a GitHub event records nothing.
- New `classifier` test pinning the capability rule, the tool enumeration, the issue carve-out, and the existence of a repo-scoped CHAT example, so a prompt fork that drops any of them fails rather than quietly restoring the over-firing.

## Docs

`spec/{03-integrations,04-event-model,05-router,10-state,11-chat}.md`, `apps/server/CLAUDE.md`, `apps/www/src/pages/docs/slack.astro`, `apps/www/src/pages/docs/workflows/answer.astro`.

## Worth a reviewer's eye

**Chat's GitHub tools have no managed-repo gate**, while the `question` route refuses unmanaged repos. Routing more questions to chat means that refusal fires less often. Pre-existing behaviour, not introduced here — but the blast radius grows slightly, so flagging it rather than folding a security change into this PR.


cliftonc:
Good catch — fixed in 7433565.

You're right that it was inconsistent with the other two writers, and the consequence is worse than "a stray row": the runner's Slack transport is the one that carries the workflow's *answer*, so a swallowed send failure meant the next chat turn confidently summarising something the user never saw. Same context drift this PR exists to prevent, arriving from the other direction.

Moved the record inside the try, after the await, as suggested.

I considered factoring the send-then-record pairing into a shared wrapper so it can't be got wrong a fourth time, and decided against it: the three senders differ in error policy on purpose — `withThreadTranscript` propagates, this one swallows-and-logs, the boot-recovery poster is a promise chain — so the wrapper would have to take the policy as a parameter and would add more than it removed. Instead the rule is now stated once in the `thread-transcript.ts` module header ("record only what was DELIVERED") and in `spec/11-chat.md`, next to the invariant it protects.

Suite still green: 2593 passed.

PR #278, opened 2026-08-06
feat(dashboard): optional "my teams' repos" filter (#169)

### LINKED ISSUE #169: GitHub team-based per-repo visibility in the admin dashboard
# GitHub team-based per-repo visibility in the admin dashboard

> **Depends on #205** — the `users` table, the verified `login` in the HMAC token, and the actor-logging columns are delivered there. This issue builds team grouping and per-repo visibility on top of that foundation. (User identity + actor logging was split out of this issue into #205.)

## Problem
Every authenticated admin sees every managed repo's workflow-runs, sessions, and stats. `authMiddleware` (`src/admin/auth.ts`) only checks the HMAC token's signature and expiry, and while #205 now carries the verified GitHub `login` in the token, nothing yet scopes what a given user sees to the repos their GitHub team actually has access to. In orgs that manage many repos this makes the dashboard noisy: a maintainer who only cares about a handful of repos still wades through the full global list.

## Users
Nearform maintainers who log into the Last Light admin dashboard via GitHub OAuth and belong to one or more GitHub org teams. Secondary: admins who log in via admin-password or Slack OAuth (no GitHub identity), and GitHub users whose team membership hasn't been synced yet — both of whom must keep seeing everything (fail-open, see below).

## Goals
- Declutter the dashboard so a GitHub-authenticated user, by default, sees only the managed repos their GitHub team(s) can access, across workflow-runs, repo-keyed stats, and sessions.
- Use native GitHub org teams and their team→repo permission grants as the visibility rule — no bespoke Last-Light grouping, config, or admin editing UI.
- Persist team / team→repo / team-membership state in SQLite, kept current via GitHub webhooks, with an admin re-sync fallback and a setup trigger.
- Keep the resolved allowed-repo set cheap to serve: resolve it from the DB on a `/me/repos` request that the UI caches with tanstack-query and filters against locally.
- Add a `repo` field to sessions so the session list can be filtered too.

## Non-goals
- **Server-side access control.** This is UI-declutter/convenience only; the server keeps returning global data on every list endpoint. Hand-crafted API calls can still see all repos. True enforcement is explicitly deferred.
- **Gating trigger/write endpoints or event dispatch** by team.
- **Per-team roles** (read vs act/trigger) or any notion beyond per-repo visibility.
- **Enumerating every org team.** Sync is scoped to teams that have a grant on a managed repo only.
- **Baking the resolved repo list into the HMAC token** (it may be too large).
- **User identity / token `login` / actor logging** — delivered by #205; consumed here.

### Identity: consume the verified login from the token (#205)
#205 extends `TokenPayload` in `src/admin/auth.ts` with the verified `login?` and threads it through `createToken` / `decodeToken`, populating it from the GitHub OAuth callback and exposing it to routes via `actorFromContext(c)`. This issue simply **reads** that `login` when resolving `/me/repos`. The full repo list is **not** stored in the token.

### Storage: new SQLite tables for teams / grants / membership
Add additive tables in `src/state/migrate.ts` (all `CREATE TABLE IF NOT EXISTS`, consistent with the existing migration style). Proposed shape (final column set to be decided during implementation):

- `github_teams (org TEXT, slug TEXT, name TEXT, PRIMARY KEY (org, slug), synced_at TEXT)`
- `github_team_repos (org TEXT, team_slug TEXT, repo TEXT, PRIMARY KEY (org, team_slug, repo))` — `repo` is the `owner/repo` full name, intersected with managed repos.
- `github_team_members (org TEXT, team_slug TEXT, login TEXT, PRIMARY KEY (org, team_slug, login))`

A new store (e.g. `src/state/team-store.ts`, mirroring `execution-store.ts` / `workflow-run-store.ts`) exposes: replace-team-repos, replace-team-members, `reposForLogin(login): string[]` (join membership → team_repos, intersect with `getManagedRepos()`), and a `lastSyncedAt()` accessor to derive the `synced` flag. Construct it in `StateDb` like the other stores (and alongside the `UserStore` from #205).

### Sync: scoped to managed-repo-granted teams
Add team methods to `GitHubClient` (`src/engine/github/github.ts`) wrapping `octokit.rest.repos.listTeams` (teams with access to a repo), `octokit.rest.teams.listReposInOrg`, and `octokit.rest.teams.listMembersInOrg`, each with `octokit.paginate`, mirroring `listInstallationRepos`. The sync routine:

1. Iterate `getManagedRepos()` (a bounded set, `src/managed-repos.ts`).
2. For each repo, list the teams with a grant → the set of relevant teams.
3. For each relevant team, pull its repos (intersected with managed) and its members.
4. Replace the DB rows for those teams. Teams with no managed-repo grant are never persisted. This is O(managed-repos), keeping the tables small even in orgs with hundreds of teams.

This requires bumping the GitHub App to org **Members: read** and subscribing to `team` / `membership` / `organization` webhook events (re-consent on install — surface as an explicit setup prerequisite).

### Keep-alive: webhooks + admin re-sync + setup trigger
In `src/connectors/github-webhook.ts`, handle `team`, `membership`, and `organization` events early (before the managed-repo action/repo filter), the way `installation` / `installation_repositories` are handled today via `handleInstallationEvent`. On these events, re-derive only the affected team's rows (or trigger a scoped re-sync for that team). Add an authenticated admin **re-sync endpoint** (e.g. `POST /admin/teams/resync`) as the fallback for orgs where the webhooks aren't wired yet, plus a setup trigger to run the initial full sync after the App is (re)installed.

### `GET /me/repos` endpoint
New Hono route in `src/admin/routes.ts`. Read the token's `login` via `actorFromContext(c)` (#205); if absent (password/Slack/`githubAllowedOrg="*"`), return the fail-open sentinel. Else call `teamStore.reposForLogin(login)`:
- If the login maps to one or more managed-repo-granting teams, return `{ repos: string[], synced: true }`.
- If it maps to **zero** teams (genuinely empty OR not-yet-synced), **fail open**: return a sentinel meaning "see everything" — e.g. `{ repos: null, synced: <bool> }` where `null`/absent means no filter. The `synced` flag is coarse (derived from `lastSyncedAt()`) and lets the UI show a hint later; both cases render fail-open now.

### Sessions: add a `repo` field
Add `repo?: string | null` to `SessionMeta` (`src/admin/sessions.ts`) and mirror it in the dashboard type (`dashboard/src/api.ts`, near the existing `repo: string | null` at ~43 / `WorkflowRun.repo` at ~100).
- **DB-backed `ChatSessionReader`** (`src/admin/chat-session-reader.ts`): source `repo` directly from `executions.repo` — surface it through `listChatThreads` / `getChatThread` in `src/state/execution-store.ts` (add `repo` to the SELECT/return shape).
- **fs-backed `SessionReader`** (`src/admin/sessions.ts`): join to `executions.repo` via the `session_id` column (`executions.session_id`, migrate.ts) or `trigger_id`. This means giving `SessionReader` access to the execution store (a new dependency to thread in at construction — to be decided during implementation).
- Sessions with no resolvable repo (repo-less chat threads) stay visible (fail-open).

### UI: tanstack-cached filter
In the dashboard (`dashboard/src/`), add an `api.ts` client call for `GET /me/repos`, cache it with tanstack-query, and apply it as a **client-side** filter on the repo-bearing views: `WorkflowList.tsx` (already renders `run.repo`), repo-keyed stats on `HomePage.tsx`, and `SessionList.tsx` (now that `SessionMeta.repo` exists). Sentinel/`repos == null` ⇒ no filter (show everything). The server continues to return global data unchanged.

## Key files to modify
| File | Change |
|------|--------|
| `src/state/migrate.ts` | Add `github_teams`, `github_team_repos`, `github_team_members` tables (additive) |
| `src/state/team-store.ts` (new) | Store for team/grant/membership rows; `reposForLogin`, `lastSyncedAt`, replace-* |
| `src/state/db.ts` | Construct the new store in `StateDb` |
| `src/state/execution-store.ts` | Surface `repo` in `listChatThreads`/`getChatThread` |
| `src/engine/github/github.ts` | Add `listRepoTeams`, `listTeamRepos`, `listTeamMembers` (paginated) |
| `src/connectors/github-webhook.ts` | Handle `team`/`membership`/`organization` events early, mirroring `handleInstallationEvent` |
| `src/admin/routes.ts` | Add `GET /me/repos` (reads `actorFromContext` from #205); add `POST /admin/teams/resync`; wire session `repo` through |
| `src/admin/sessions.ts` | Add `repo` to `SessionMeta`; source it in `SessionReader` via executions join |
| `src/admin/chat-session-reader.ts` | Populate `repo` from `executions.repo` |
| A sync module (new, e.g. `src/engine/github/team-sync.ts`) | Managed-repo-scoped sync routine + setup trigger |
| `dashboard/src/api.ts` | `ManagedRepos`-style type + `GET /me/repos` client; `repo` on session type |
| `dashboard/src/components/WorkflowList.tsx`, `SessionList.tsx`, `HomePage.tsx` | Client-side repo filter from cached `/me/repos` |
| GitHub App config / `.env.example` | Document `Members: read` + new webhook subscriptions as setup prerequisite |

## Decisions made during exploration
- Visibility model → native GitHub teams + team→repo grants, DB-backed (GitHub already models access; avoids per-request fan-out; orgs may have many teams).
- Storage/refresh → SQLite tables kept live by `team`/`membership`/`organization` webhooks, plus admin re-sync + setup trigger fallback; UI tanstack-caches `/me/repos` (client-side, not a heavy server-side cache).
- Sync scope → only teams with a grant on a managed repo: iterate `getManagedRepos()`, list teams-with-access per repo, pull members for those teams only — O(managed-repos), small tables.
- GitHub App → bump to org `Members: read` and subscribe to `team`/`membership`/`organization`; explicit setup prerequisite requiring re-consent.
- Enforcement → UI-declutter/convenience only, client-side filtering; server keeps returning global data. Server-side access control deferred.
- Empty-membership semantics → fail OPEN: zero managed-repo-granting teams ⇒ sentinel "see everything", consistent with password/Slack/`allowedOrg="*"` super-admin. `/me/repos` returns allowed set + coarse `synced: boolean`.
- Identity in token → the verified `login` is carried by #205 (do NOT bake the full repo list into the token).
- `/me/repos` → read token `login`, look up DB memberships → intersect with `getManagedRepos()`, return `{ repos, synced }`; UI caches and filters locally.
- Session repo → add `repo` to `SessionMeta`; DB-backed reader reads `executions.repo`, fs-backed reader joins via `session_id`/`trigger_id`; repo-less sessions stay visible.
- Filtered views → workflow-runs, repo-keyed stats, AND sessions; sentinel/unsynced ⇒ no filter.

## Acceptance criteria
- A GitHub-authenticated user who belongs to team(s) granting access to a subset of managed repos sees only those repos' workflow-runs, repo-keyed stats, and sessions in the dashboard by default.
- `GET /me/repos` returns `{ repos, synced }` — a concrete list for a synced user with team grants, and a fail-open sentinel (no filter) for password/Slack/`allowedOrg="*"` users and for GitHub users with zero managed-repo-granting teams.
- Team, team→repo, and team-membership rows exist only for teams with a grant on a managed repo; `team`/`membership`/`organization` webhooks keep them current; the admin re-sync endpoint reproduces the same state on demand.
- `SessionMeta` (server and dashboard type) carries `repo`; the session list is filtered by the same allowed-repo set, and repo-less sessions remain visible.
- The server's list endpoints (`/workflow-runs`, `/sessions`, `/stats`) still return global data — filtering is client-side only.
- Store and route behavior is covered by Vitest tests following the existing `tests/` store/route patterns (raw `Database` + `migrate()`, faked `GitHubClient`).

## Open questions
- Exact `/me/repos` sentinel shape (`repos: null` vs absent vs explicit `unfiltered: true`) — to be decided during implementation.
- Whether the fs-backed `SessionReader` gets the execution store injected at construction or resolves repo via a shared helper — to be decided during implementation.
- How the `organization` webhook (member added/removed from org) should map to team-membership rows vs relying on `membership` events — to be decided during implementation.
- Whether the setup trigger is a boot-time hook, an install webhook handler, or purely the admin re-sync endpoint — to be decided during implementation.
- Whether to add a DB index on `github_team_members(login)` for `reposForLogin` — likely yes; confirm during implementation.

## Out of scope
- Server-side enforcement / access control on any endpoint (deferred to a larger later effort).
- Per-team roles (read vs trigger/act) and any write-path gating.
- Enumerating or storing teams that have no managed-repo grant.
- Baking the resolved repo list into the token.
- A UI for editing team→repo visibility (GitHub remains the source of truth).
- User identity records, token `login`, and actor logging (delivered by #205).

---
*Spec generated from a Slack thread via Last Light explore workflow. User identity + actor logging split out into #205.*

cliftonc:
Closes #169.

Adds an **optional, opt-in** header filter that narrows every dashboard list to the managed repos a person's GitHub org teams own. Plus the plumbing that turned out to be needed to make any repo filtering work at all.

## What a dry run against the live instances changed

Before enabling this anywhere, I resolved it for real against `nearform` and `drizby`:

| | |
|---|---|
| `cliftonc` @ nearform, 8 managed repos | 1 team (`nearformers`) → **4 of 8** |
| The 4 it would have hidden | `lastlight`, `mac`, `td-toolkit`, `lastlight-flue` |
| Permission on all 4 | **`admin`**, via org ownership — no team grant |
| drizby | `mirevue` → 0 teams; `cliftonc` is a personal account → nothing to scope by |

So the rule the issue specified — "use team→repo permission grants as the visibility rule" — **under-approximates access**. Org owners and direct collaborators reach repos no team grants them. As a *default*, this feature would have hidden the Last Light repo from the person who writes Last Light.

Underneath is a conflation worth naming: **"what can you reach" (permission) vs "what do you work on" (involvement)**. They coincide for a regular org member and diverge completely for an owner — for whom no access-derived rule can declutter anything, because they can reach everything.

**The resolution is to invert who chooses it.** The narrowing is now opt-in and off by default, remembered per browser. That makes the imperfect rule correct rather than merely tolerable: as a default it hides your own work; as a filter you switched on, "the repos my teams own" is precisely the decluttering you wanted, and one click undoes it. Nothing is applied to anyone who didn't ask, and it is never described as visibility or access.

## Two bugs the dry run exposed

- **A personal-account owner was treated as a failure.** `organization(login:)` on a user account answers `{ organization: null }` *plus* a NOT_FOUND error, and Octokit throws whenever `errors` is present — so the null check meant to skip it was unreachable. drizby would have logged a warning per resolution.
- **One org failing while another succeeded still produced a filter.** If `nearform` resolved and `mirevue` errored, the result narrowed to nearform's repos and hid every mirevue repo. A partial org set is a partial answer; any failure now fails open, matching truncation.

Both have tests; neither did before.

## The repo-name bug (found chasing "why not `org/repo` everywhere?")

`workflow_runs.repo` stores a **bare** name with the account in a separate `owner` column; `executions.repo` is qualified from the dispatcher but bare from the phase executor — which is every workflow run. `/me/repos` returns qualified `owner/repo`. So the first version of the filter compared `lastlight` against `{nearform/lastlight}`, never matched, and **a non-null non-match hides the row**. Fixed by normalizing on the read side where the owner is reachable, returning NULL — "always visible" — when a name can't be qualified. Storage really should just be `owner/repo` everywhere: that's #279, kept separate because it touches Octokit call sites where a mistake means commenting on the wrong repo.

## What's here

| Layer | Change |
|---|---|
| `src/state/team-store.ts` (new) + `migrate.ts` | Four additive tables. A **cache, not a mirror** — rows exist only for teams of somebody who opted in, so absence means "unknown", never "no access". |
| `src/engine/github/team-visibility.ts` (new) | Lazy per-user resolver: budgets, in-flight dedup, stale-while-revalidate, early exit once a team's grant covers every managed repo. Every failure path fails open. |
| `src/engine/github/github.ts` | `listUserTeams` / `listTeamRepos` via GraphQL, bounded and self-metering. `Organization.teams(userLogins:)` does the membership intersection server-side, so cost tracks the *user's* team count, not the org's repo count — 3000 managed repos + one team granting 250 costs 4 requests (tested). |
| `src/connectors/github-webhook.ts` | `team`/`membership`/`organization` **invalidate** rather than re-derive; re-deriving would put an unbounded org walk on the delivery path. |
| `src/admin/routes.ts` | `GET /me/repos`, `POST /me/repos/resync` (**self only** — see review), and `?repos=` on `/workflow-runs`. |
| `src/state/execution-store.ts`, `workflow-run-store.ts` | Repo-name qualification on both read paths. |
| `src/admin/sessions.ts`, `chat-session-reader.ts` | `SessionMeta.repo`, via an injected lookup rather than a `StateDb` dependency. |
| `dashboard/` | `useVisibleRepos` (no tanstack-query here, so a module-scoped shared cache) + the header control + `?repos=` on the run lists. |

## Review feedback

`?login=` is gone from resync. Beyond the reported issue: a password-only session has no GitHub identity to check an override *against*, and the response's `teams` field discloses membership of teams that may be secret.

## Verification

Full gate green — 16/16 turbo tasks, 2749 server tests, dashboard typecheck, `astro check` 0 errors. Resolution was additionally validated against live GitHub data for `cliftonc` across `nearform`, `mirevue` and the personal account.

**Not verified:** I haven't run the dashboard and looked at the header control. It typechecks and the logic is tested, but spacing and dark-mode contrast are unexamined.


cliftonc:
Thanks — the `?login=` finding is valid and I've taken the suggested fix in 8c86b86. Resync is now always self, with no override param.

The reasoning I'd add to the report: it isn't only that the caller might lack elevated standing, it's that there's *nothing to check standing against*. A password-only session has no GitHub identity at all, so `actorFromContext` returns undefined and any guard comparing the param to the session has no left-hand side. And the disclosure is worse than "which repos" — `teams` names the org teams somebody belongs to, and GitHub teams can be secret. The debug affordance didn't justify either, so I dropped the override rather than gating it.

Two other things landed in the same commit, both found while chasing your comment:

**The filter was broken, in the hide direction.** `workflow_runs.repo` stores a *bare* repo name with the account in a separate `owner` column, and `executions.repo` is qualified when the dispatcher writes it but bare when the phase executor does — which is every workflow run. `/me/repos` returns qualified `owner/repo`. So the comparison never matched, and a non-null non-match hides the row: the feature would have blanked the dashboard for anyone it was enabled for. That's precisely the outcome the fail-open design is meant to make impossible — the resolver's property was correct, the join key on the other end wasn't. Fixed by normalizing on the read side where the owner is available, yielding NULL (= "always visible") when a name can't be qualified. Regression tests pin the bare case on both paths.

**The scope moved to the server.** The home panels were over-fetching 40 rows to show 5, which was a workaround for client-side filtering and left `total` as the global count. `?repos=` is the plural sibling of the `?repo=` param the Repos tab already uses — a caller-supplied query filter, not enforcement — so the lists now ask for exactly what they render and paging stays honest. Plus a global "my repos / all repos" toggle in the header, so there's always a way out of the narrowing.

Storage should just be `owner/repo` everywhere; that's tracked separately rather than smuggled into this PR.

PR #288, opened 2026-08-07
fix(dashboard): show the repo-scope toggle whenever teamVisibility is on (#287)

### LINKED ISSUE #287: The "my teams' repos" toggle hides itself in the one state where its re-sync is needed
Found while enabling `teamVisibility` on a live instance.

## The bug

`RepoScopeToggle` (`apps/server/dashboard/src/components/StatsHeader.tsx`) returns `null` when `!canScope && !degraded`, and `canScope` is `Boolean(meta?.repos && meta.repos.length > 0)` — i.e. the control renders **only once real team grants resolved**.

That hides it in precisely the state a user needs it. The common path in: somebody enables the feature, sees no toggle, creates a team and grants it repos — and still sees nothing, because the "no teams" answer is cached for `ttlMinutes` (60 by default).

Worse, `visibleRepos()` is **stale-while-revalidate**: after the TTL lapses the *first* load still returns the stale answer and only kicks the refresh off behind it. So even waiting an hour takes two page loads to show a change.

## `degraded` is dead code

```ts
degraded: scope === "mine" && allowed === null,
```

Opting into `mine` requires the toggle. The toggle is hidden whenever `!canScope && !degraded`. So in every fresh fail-open state `degraded` is unreachable — the branch written to explain a failed resolution can only be reached by a resolution that *succeeded and later broke*. The "my teams (unavailable)" label and its tooltip are effectively unshippable today.

## The escape hatch exists but nothing calls it

`POST /admin/api/me/repos/resync` is implemented, `api.meReposResync` wraps it, and `useVisibleRepos` exposes `resync()`. Grepping the SPA, **no component calls it.** The only route back is the TTL.

## Fix

Gate on the **operator's switch**, not on whether grants resolved: render the control whenever `teamVisibility.enabled` is on, and have the unresolved states explain themselves and carry the re-sync.

`teamVisibility.enabled: false` stays the one hidden state — and is the one a re-sync provably cannot change, since `resync()` short-circuits on `config.enabled` before touching GitHub. So the client's render rule and the server's guard agree by construction rather than by coincidence.

Since `reason` has nine values and the mapping is the part that rots, the decision should be a pure function pinned by a test against the server's own `VisibilityReason` union, so adding a reason without deciding how the UI answers it fails the build.

Dashboard-only — no server change — but it still needs a release to reach prod, since the SPA ships inside `lastlight-core`.

cliftonc:
Closes #287.

The "my teams' repos" control rendered only once real grants resolved, which hid it in the one state where it's needed — and hid the re-sync that fixes that state.

## What was wrong

Enable `teamVisibility`, see no toggle, create a team and grant it repos — and still see nothing, because the "no teams" answer is cached for `ttlMinutes` (60 by default). `visibleRepos()` is also **stale-while-revalidate**, so the first load after expiry serves the old answer and only refreshes behind it: two loads to see a change.

`POST /me/repos/resync`, `api.meReposResync` and the hook's `resync()` all existed. **Nothing in the SPA called any of them.**

And `degraded` was dead code:

```ts
degraded: scope === "mine" && allowed === null
```

Opting into `mine` requires the toggle; the toggle was hidden whenever `!canScope && !degraded`. The branch written to explain a failed resolution was reachable only via a resolution that succeeded and later broke.

## The rule

Gate on the **operator's switch**, not on whether grants resolved.

`teamVisibility.enabled: false` remains the one state it isn't drawn in — and is the only state a re-sync provably cannot change, since `resync()` short-circuits on `config.enabled` before touching GitHub. So the client's render rule and the server's guard now agree by construction, not by coincidence.

| `reason` | control |
|---|---|
| `disabled` | hidden — the operator turned it off |
| `no-teams`, `no-identity`, `unavailable`, `too-many-teams`, `truncated`, `budget`, `error` | shown, explanatory, click to re-sync |
| `ok` with grants | shown, toggles the filter (unchanged) |

In the unresolved state the button names the reason and clicking it re-asks the server, rather than toggling a scope that would narrow nothing — the control *is* the escape hatch.

## Not changed

Still opt-in, still off by default, still remembered per browser, every failure path still fails **open**. Nothing narrows that didn't narrow before — this only makes an existing dead-end state legible and recoverable.

## Testing

`scopeControlState` is extracted as a pure function so the mapping is testable without React (the dashboard has no test runner; core's suite imports its plain-TS modules directly, as `dashboard-repo-links.test.ts` does next door). The expectation table is typed as the **server's** `VisibilityReason`, so adding a reason without deciding how the UI answers it is a build error.

13 new tests. `pnpm turbo run typecheck test` 16/16 green; `astro check` 0 errors.

Docs updated per `docs-sync`: `spec/02-configuration.md`, `spec/03-integrations.md`, and the site's configuration page — including a "just created a team?" note explaining the TTL and the two-load staleness, which is the thing that actually confuses people.

⚠️ Dashboard-only, no server change — but the SPA ships inside `lastlight-core`, so it still needs a release to reach a deployment.


cliftonc:
Second commit pushed — the filter now counts **repos you own**, not just your teams'.

Teams are an *org* concept, so a repo under a personal account can never be team-granted. A strictly team-derived filter therefore hid every personal repo an instance manages — nine of eleven on drizby. That gap was pure artefact rather than the deliberate approximation the rest of the feature makes: team grants stand in for involvement, and nobody is more involved in a repo than its owner.

### The rule is `owner === login`

Not "the owner is not an organization". Those two pick out the same set on a single-owner instance and the second is a **disclosure bug** on any other — it would put everybody else's personal repos into your filter. GitHub shares one namespace between users and orgs, so an exact login match is unambiguous, and it costs no API call (the owner is already in the managed string). There's a test pinning that case specifically, since the two readings are indistinguishable on the instance this was developed against.

### Two properties worth reviewing

**Derived per request, never persisted.** It falls out of `(login, live managed list)`, both already in hand at projection — the same place team repos are re-intersected with that list. So a de-managed repo leaves both halves at once, and the owned half has no cache of its own to go stale.

**An incomplete team answer is deliberately NOT rescued by it.** `truncated` / `error` still fail open. "Your own repos plus an unknown fraction of your teams'" is still a partial answer, and narrowing to it would confidently hide exactly the org repos the failed half would have contributed — the same reasoning that already makes a partial org set fail the whole pass. Only a complete pass (`ok` or `empty`) is augmented, which is also what lets `empty` produce a filter now instead of always failing open.

Direction of safety is unchanged: this only ever **adds** to the allow-list, so it can only make the filter less narrowing.

### Label

"my teams" → "my repos", since the set is no longer team-derived. Carried through `config/default.yaml`, `spec/02-configuration.md`, `CLAUDE.md` and the site.

### Verification

6 new resolver tests (28 total in that file, so the 22 pre-existing still pass — worth noting since `fromCache`'s status branching changed). Full gate 16/16, `astro check` 0 errors.

### Note for drizby

With this, `owned(9 cliftonc) ∪ teams(mirevue)` = every managed repo, so the filter is a no-op on that instance. That's the honest answer — you genuinely are involved in all of them — but the feature only earns its keep somewhere org-owned like nearform.

PR #292, opened 2026-08-07
feat(github): publish signed commits via createCommitOnBranch

### LINKED ISSUE #268: Sandbox-built commits are unsigned, so base merges block PRs on repos that require signed commits
On a repo with GitHub's `required_signatures` rule enabled, a fix run that merges the base branch does not unblock the PR — it blocks it permanently, and no subsequent fix run can clear it.

## What happens

`workflows/prompts/dependabot-ci-fix.md` instructs the agent to run `git fetch origin <base>` then `git merge --no-edit FETCH_HEAD` and push. Any commit built by `git` in the sandbox arrives unsigned.

`required_signatures` is evaluated over **every commit reachable in the PR**, not just the head. So one unsigned commit anywhere in the branch is terminal. The observed symptom is a PR sitting at `mergeStateStatus: BLOCKED` with `mergeable: MERGEABLE`, every required check green and zero required approvals — the only cause being a commit with:

```
verification: { verified: false, reason: "unsigned" }
```

Dependabot's own commits on the same branch are `verified: true`, because anything pushed through GitHub's API is signed with the `web-flow` key automatically. Only ours are unsigned.

Recovery requires `@dependabot recreate` or a human. Recreate discards the PR's CI history and regenerates the lockfile, which re-rolls dependency resolution — I've seen that turn a fully green PR red (see the companion issue on time-gated failures).

## Why the App token does not help

These are unrelated mechanisms and easy to conflate:

| | What it does | Where it comes from |
|---|---|---|
| **Token** | authenticates the *push* | GitHub App installation |
| **Signing key** | signs the *commit object* | `commit.gpgsign` + a private key on the machine |

The sandbox has the first and not the second. No amount of write permission makes a locally-built commit signed. There is no commit signing configured anywhere in the repo today — the only `gpgsign` settings are `tag.gpgsign` for releases.

Note the merge itself is correct and deliberate: it makes a `behind` PR current and regenerates lockfiles on a `dirty` conflict. The defect is the mechanism, not the intent.

## Proposed fix

Use `PUT /repos/{owner}/{repo}/pulls/{number}/update-branch` instead of a sandbox `git merge` whenever the goal is only "make this PR current". GitHub performs the merge server-side, so the commit is committed by `GitHub <noreply@github.com>` and signed with the `web-flow` key — the same path as the UI's "Update branch" button. No key management.

Where the goal is only to re-run CI after a transient failure, create no commit at all: `POST /repos/{owner}/{repo}/actions/runs/{run_id}/rerun-failed-jobs`.

### Verified

GitHub's REST reference for `update-branch` doesn't explicitly document the signing behaviour of the commit it creates, so I checked rather than assumed. Run against two PRs on a repo enforcing `required_signatures`, both resulting merge commits came back:

```
committer=GitHub <noreply@github.com>  verified=true  reason=valid
```

Useful tell for auditing history after the fact: `update-branch` produces `Merge branch 'main' into <branch>`, whereas a sandbox merge produces `Merge remote-tracking branch 'origin/main' into <branch>` (or `Merge branch 'main' of <url> into <branch>` if it went via `git pull`). The two mechanisms are distinguishable from the commit message alone.

## What this does not cover

`update-branch` cannot resolve conflicts; it fails on a `dirty` PR. Lockfile regeneration still needs a sandbox, and any commit produced there is still unsigned. On signature-enforcing repos those PRs still can't be landed by lastlight — narrower than today's gap, where *every* base merge is terminal rather than just the conflicted ones, but not zero.

## Open question

**Should lastlight hold a signing key?** An SSH or GPG key in the sandbox with `commit.gpgsign` would make every path work, including conflict resolution. The cost is a key to mint, store and rotate, plus a signing identity registered against the App's account for GitHub to mark it verified. Probably not worth it if `update-branch` covers the common case — but it's the answer if the `dirty`-PR gap bites in practice.

robinbowes:
Closes #268.

On a repo with GitHub's `required_signatures` rule, every commit Last Light
produced was unsigned and therefore blocked. Every code-writing workflow ended
by running `git add -A && git commit && git push` in the sandbox, and a commit
object built by `git` there carries no signature — the token authenticates the
*push*, while a signature is a property of the *commit object*. One unsigned
commit anywhere in a branch blocks the PR permanently, and no later run clears
it.

This stops building the published commit locally. GitHub builds and signs it.

## What this does

A new Pi tool, **`github_publish`**, replaces `git add && git commit &&
git push` as the publish step in all nine code-writing prompts. It diffs the
working tree against the branch's current remote tip and hands the change set
to GitHub's GraphQL **`createCommitOnBranch`** mutation, which creates the
commit server-side and signs it. Local `git commit`s remain scratch working
state that never leaves the sandbox — they are folded into the published
commit. No signing key exists anywhere; the only credential is still the
short-lived installation token.

**Correcting the spec.** #268's spec proposed generalizing
`GitHubClient.pushFiles()` — the REST Git Data API — on the premise that
GitHub signs at `createCommit` time. It does not: `signature` there is an input
you supply, so that path produces unsigned commits today. Measured side by side
on one repo with one token: REST `git/commits` → `verified: false, reason:
"unsigned"`; GraphQL `createCommitOnBranch` → `verified: true, reason: "valid"`.
Detail in [`docs/plans/signed-commit-publish/README.md`](docs/plans/signed-commit-publish/README.md)
and the [comment on #268](https://github.com/nearform/lastlight/issues/268#issuecomment-5216102804).

## Behaviour worth reviewing

**It refuses rather than publishing something wrong.** `FileAddition` carries
only `path` and base64 `contents` — no mode. So a new executable file, a
symlink, a submodule pointer, or a mode change on an existing file cannot be
expressed, and the tool refuses before any remote write, naming the files. A
content-only change to a file that is *already* `100755` is fine: GitHub keeps
the base tree's mode (measured). The refusal names the usual escape — make the
file non-executable and invoke it through its interpreter — and
`skills/building/SKILL.md` says the same.

**It refuses when the checkout is behind the branch.** The change set is the
working tree measured against the remote tip, so publishing from a stale
checkout would record another party's files as deletions. `expectedHeadOid`
does not catch this, because the tip has not moved since we read it. The tool
asserts `git merge-base --is-ancestor <tip> HEAD` and refuses, naming a
recovery for each cause — `reset --mixed` when the tip is our own earlier
publish, commit-then-merge when somebody else pushed.

**It checks its own work.** The mutation returns the commit's signature, so
every publish asserts GitHub actually signed it and fails loudly otherwise,
including when `signature` comes back `null` — which is how GraphQL reports an
unsigned commit.

**No retry, deliberately.** `expectedHeadOid` is required by the schema, so a
tip that moved after the read is rejected. Re-reading and re-diffing would
render the other party's added files as deletions, so a retry is not safe.
`STALE_DATA` is named in the error, because GitHub's own REST read lags its
GraphQL write and can produce it with nobody else pushing.

**`include` / `exclude`.** The five artifact-only phases pass
`include: [".lastlight"]` so they publish exactly what `git add .lastlight/`
used to, rather than the whole tree.

**`github_push_files` is removed.** REST Git Data API, produces unsigned
commits, zero callers, hardcodes `mode: "100644"`, cannot express deletions —
strictly weaker than `github_publish`, and most reachable exactly when
`github_publish` refuses. `github_create_or_update_file` (Contents API, a
different path) is unchanged.

## Open

- **`[bot]` attribution is unverified.** The probes used a user PAT and
  returned `GitHub <noreply@github.com>`. Under an installation token the
  committer is expected to be the App's `[bot]` identity, but this branch does
  not demonstrate it. `spec/09-sandbox.md` and `apps/server/CLAUDE.md` are
  hedged accordingly. Closing it needs one workflow phase run end to end.
- The `dirty`-PR case #268 raised is handled by the default mechanism: the
  change set is read from the working tree regardless of how it got there, so a
  conflict-resolved, lockfile-regenerated tree publishes like any other.
  `pulls.updateBranch` is not required.

## Testing

`pnpm turbo run typecheck test build` — 21/21 tasks, 2937 tests. agentic-pi
277 unit tests. agentic-pi's JSONL fixtures are byte-identical.

New coverage: change-set computation against real temp git repos (modes,
binary, deletions, renames-pinning, include/exclude); the mutation against a
fake GitHub; the tool's refusal paths asserting that *no* HTTP request reaches
GitHub; signature assertion including the `null` case; and a guard test over
`workflows/prompts/`, `agent-context/` and `skills/` that fails if a packaged
prompt reintroduces `git push`. That guard catches the literal invocation only
— prose instructing a push without naming the command is not machine-checkable,
and the test documents that limit.

https://claude.ai/code/session_01AVpW8ddTZKTTLTHtM6zHwB


robinbowes:
Follow-ups filed for the two items left open above:

- #293 — verify `[bot]` attribution end to end under an App installation token, and measure whether `github_create_or_update_file`'s Contents-API path is signed. Until the first lands, `spec/09-sandbox.md` and `apps/server/CLAUDE.md` stay hedged.
- #294 — robustness and test-coverage debt in the publish path (the untested base fetch, the shallow-clone depth, coverage gaps, cosmetics).

cliftonc:
Thanks @robinbowes - I will get this deployed to my private instance.  I will also share an invite to the org that manages it, so you can login to it via github and test it with a real running instance (https://lastlight.drizby.com/admin).

Issue #322, opened 2026-08-11
Cancel a PR review superseded by a new push, instead of discarding its output


robinbowes:
## Problem

A push that lands while a PR-scoped run is in flight loses the run lock. `runLockDrop` (`src/engine/pr-decisions.ts:303`) drops the **new** event with `run-in-flight: pr-review <runId> is already working this PR`, so the in-flight run keeps reviewing a tree that no longer exists. When it finishes, `staleAgainstCurrentHead` (`src/workflows/handlers/post-review.ts:304`) correctly refuses to post the result.

The review is suppressed. The tokens are not. The 30-minute `check-prs-awaiting-review` sweep is the re-pickup.

So the gap is narrow: **we already detect this, we just detect it after paying for it.** The proposal is to move the detection to the moment the push arrives — cancel the superseded run instead of letting it finish and discarding the output.

## Evidence — this is robustness, not a cost fix

Measured on a homelab instance running v0.25.5, 11 Aug 2026, across the 44 `pr-review` runs that completed that day ($10.00 total):

| Category | Runs | Cost |
|---|---:|---:|
| Superseded **mid-run** — what this would cancel | **0** | **$0.00** |
| Stale at dispatch | 0 | $0.00 |
| Reviewed the then-current head, finished before the next push | 44 | $10.00 |

That day was the adversarial case: eight PRs opened within 38 seconds by a scripted rollout, then bulk force-pushed four to five times over four hours, 3–6 reviews per PR. **Not one run was overtaken** — reviews took 1–10 minutes and pushes were 20–50 minutes apart.

I'd rather file that honestly than oversell the motive. The window scales with review duration, so a repo whose reviews take 10+ minutes on a rapidly-pushed PR is inside it continuously; but on this data it is a correctness-under-adverse-timing feature, not a spend fix.

**Method** (so this can be re-run before prioritising): take each run's `context.prState.headSha` and window from `workflow_runs`, the head's committer date from `GET /repos/{r}/commits/{sha}`, and every `head_ref_force_pushed` event from `GET /repos/{r}/issues/{n}/timeline`; a run is superseded when the first push strictly newer than its head lands before `finished_at`. One trap worth knowing: **the force-push event fires 5–10s *after* the committer date of the commit it delivers**, so a naive comparison flags every run as stale. Require a real gap (>60s) or match the pair — this cost me three false positives.

## Proposed scope

[...]

PR #324, opened 2026-08-11
docs(review): spec cancel-on-push for superseded PR reviews

robinbowes:
Adds `docs/plans/review-cancel-on-push/README.md` — a plan doc for cancelling a PR-scoped run that a new push has made worthless, instead of letting it finish and discarding the output. Documentation only; no code.

Specs the issue filed as #322.

## What the gap is

A push landing mid-run loses the run lock: `runLockDrop` (`src/engine/pr-decisions.ts:303`) drops the **new** event, so the in-flight run keeps reviewing a tree that no longer exists, and `staleAgainstCurrentHead` (`src/workflows/handlers/post-review.ts:304`) then correctly refuses to post it. The review is suppressed; the tokens are not. So the detection already exists — it just runs after the spend.

## What the doc contains

- **The measurement, stated against the proposal.** 0 of 44 `pr-review` runs were overtaken on a homelab instance's worst day for it (eight PRs bulk-force-pushed 4–5 times over four hours). Filed as robustness under adverse timing, not a spend fix. The method is written out so it can be re-run before anyone prioritises the work — including the trap that produced three false positives for me: the `head_ref_force_pushed` event fires 5–10s *after* the committer date of the commit it delivers, so a naive comparison reports every run as stale.
- **Scope as an opt-in `cancel_on_push:` key**, derived like `pr_scoped:` so remapping `routes.github.pr_review` to a fork can't silently inherit or lose it. `pr-review` only; the fix family and the merge workflow are argued out individually.
- **Two guards, with the second one's justification.** Cancel only when the head really moved *and* the delta is material. Cancelling and re-dispatching are not one action — a generated-only push is suppressed by the existing gate, so an unguarded cancel leaves the material change with **no** review at all. That is `staleAgainstCurrentHead`'s own *"dropping a review is only acceptable when a replacement is guaranteed"* read forwards.
- **A five-phase implementation sketch**, the first phase of which is extracting the admin route's cancel body (`src/admin/routes.ts:1652`) into something reusable — a second caller reimplementing part of it would strand an `in_progress` check run or leak a sandbox pod.
- Non-goals, testing, and two open questions.

## Deliberately unresolved

Open question 2 — whether the supersede path's extra `getChangedPathsBetween` call is worth it versus cancelling on any head move — is left open rather than decided. The evidence says the call is rare by construction; the guard it powers is the one keeping a material change from going unreviewed. My recommendation is in the doc, but it's the kind of trade worth settling in review before any code exists.

https://claude.ai/code/session_017csqMy58ivRbLTKojhQnpJ

### DIFF
--- a/docs/plans/review-cancel-on-push/README.md
+++ b/docs/plans/review-cancel-on-push/README.md
@@ -1,0 +1,223 @@
+# Cancel-on-push — implementation plan
+
+Kill a PR-scoped run that a new push has made worthless, instead of letting it
+finish and throwing the result away.
+
+Today the run lock resolves a push-during-a-run the other way round: the
+**new** event loses. `runLockDrop` (`src/engine/pr-decisions.ts:303`) drops it
+with `run-in-flight: pr-review <runId> is already working this PR`, the
+in-flight run reviews a tree that no longer exists, and `post-review`'s
+`staleAgainstCurrentHead` (`src/workflows/handlers/post-review.ts:304`)
+correctly refuses to post it. The review is suppressed; the tokens are not.
+The 30-minute `check-prs-awaiting-review` sweep is the re-pickup.
+
+So the gap is narrow and precise: **we already detect this, we just detect it
+after paying for it.** This plan moves the detection to the moment the push
+arrives.
+
+## Evidence — and why this is robustness, not a cost fix
+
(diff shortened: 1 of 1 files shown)


cliftonc:
Looks like a great addition 👍

PR #330, opened 2026-08-11
fix(dashboard): share one status palette, repairing the feedback CVD pair

### LINKED ISSUE #329: Dashboard status colours are defined twice and the feedback pair fails CVD separation
Found while checking #328 for colour clashes with the dashboard's other charts. Two problems, one fix.

## 1. The feedback chart's dark palette fails colour-vision separation

`FeedbackPage.tsx` renders positive and negative feedback counts as adjacent bars:

```ts
const CHART_DARK = { positive: "#86efac", negative: "#fca5a5", … };
```

That pair measures **ΔE 5.8 for deuteranopes** (OKLab ×100) — below the ΔE 6 floor. Positive and negative feedback are the most semantically opposed quantities in the view, and a deuteranope cannot reliably tell those two bars apart. This is the same defect #328 fixes on the Home page, where the identical pastels were in use.

## 2. The dashboard now has two greens for "good" and two reds for "bad"

After #328 the two pages disagree:

| meaning | Home page | Feedback page | ΔE apart |
|---|---|---|---:|
| good | `#0ca30c` | `#07a06f` | 9.3 |
| bad | `#cc2b5e` | `#dc2626` | 7.3 |

That is the worst possible distance — near enough to read as "the same idea", far enough to look like a mistake, and it invites a reader to wonder what the difference signifies when it signifies nothing. Both pages also carry their own duplicated `CHART_DARK` / `CHART_LIGHT` literals, with a comment in `FeedbackPage.tsx` noting they are the "same values as HomePage's charts" — a copy that is now false.

Note the Home page's *categorical* series (token input/output/cache, cost) are fine and were checked: every cross-pair against the new status colours is ΔE 12.9–44.6, and no status hue is reused as a series colour.

## Proposed fix

Lift the status palette into one module the whole SPA imports — `src/lib/status-colors.ts` — keyed by meaning rather than by page:

```ts
export const STATUS = {
  good:    "#0ca30c",  // succeeded · positive
  neutral: "#6b7280",  // skipped (hatch)
  info:    "#4a7fb5",  // deferred
  bad:     "#cc2b5e",  // failed · negative
};
```

Home page maps its four outcomes onto it; Feedback maps `positive`/`negative`. That makes the app consistent **and** repairs the CVD failure in one move, because the shared pair is the validated one (green↔red ΔE 8.7, clearing the target in both themes).

This is the same "derive once" argument #328 makes about `executions.success`: one concept with two definitions drifts, and the drift is invisible until someone measures it.

Worth pinning with a source-level test in the manner of `tests/admin/dashboard-config-mirror.test.ts`, so a future page cannot quietly reintroduce its own status hexes.

robinbowes:
Closes #329.

## Two problems, one fix

**1. The feedback chart failed colour-vision separation.** `FeedbackPage`'s dark pair `#86efac` / `#fca5a5` measures **ΔE 5.8 for deuteranopes** (OKLab ×100), below the ΔE 6 floor — on positive vs negative feedback counts, the two most semantically opposed quantities in the view.

**2. The dashboard had two greens for "good" and two reds for "bad".** After #328 the pages disagreed by ΔE 9.3 and 7.3 — the worst possible distance: near enough to read as the same idea, far enough to look like a mistake.

Both come from the same cause, which is the one #328 makes about `executions.success`: **one concept with two definitions drifts, and the drift is invisible until someone measures it.** `FeedbackPage` even carried a comment claiming its palette held "the same values as HomePage's charts" — true when written, false by the time anyone read it.

## The fix

`dashboard/src/lib/status-colors.ts` is now the only place in the SPA that names a status colour, keyed by **meaning** rather than by page:

```ts
export const STATUS = {
  good:    "#0ca30c",  // succeeded · positive
  neutral: "#6b7280",  // skipped (hatch)
  info:    "#4a7fb5",  // deferred
  bad:     "#cc2b5e",  // failed · negative
} as const;
```

Home maps its four outcomes onto it; Feedback maps positive/negative. Repairing the CVD failure falls out of removing the duplication, because the shared pair is the validated one — every pair clears ΔE 8 CVD and the 15 normal-vision floor in both themes.

`score` stays local to Feedback: it is a **categorical** line colour, not a status, and the distinction is the whole reason status colours are reserved.

Also removes `CHART.success` / `CHART.error` from `HomePage` — the bars moved to `OUTCOME` in #328 and neither key has had a reader since.

## No clash with the other charts

Checked before filing #329. Every cross-pair between the status palette and the categorical series it sits beside (token input/output/cache, cost) measures ΔE 12.9–44.6, and no status hue is reused as a series colour anywhere.

## The test

`tests/admin/dashboard-status-palette.test.ts`, following `dashboard-config-mirror`: the SPA is a separate Vite app with no import edge to core, so "a page hardcoded its own status hex" is a fact about text, not types.

It scans source **with comments stripped**, which is deliberate. `FeedbackPage` names the retired `#86efac` / `#fca5a5` on purpose, to record why they went — a test that forbade saying so would push the explanation out of the codebase and leave the bug's epitaph nowhere. The invariant is about what the code *uses*.

Written first and watched fail 7/7.

## Verification

- Full server suite **2986 passed**, 20 skipped, 0 failed (up 7 from the new file),
  re-run after rebasing onto current `main`.
- `@lastlight/dashboard` typechecks clean.
- Workspace `turbo run typecheck` clean, dep-cruiser included.

https://claude.ai/code/session_017csqMy58ivRbLTKojhQnpJ

### DIFF
--- a/apps/server/dashboard/src/components/FeedbackPage.tsx
+++ b/apps/server/dashboard/src/components/FeedbackPage.tsx
@@ -18,6 +18,7 @@ import {
   type FeedbackSummaryRow,
 } from "../api";
 import { useTheme } from "../hooks/useTheme";
+import { STATUS } from "../lib/status-colors";
 
 /**
  * Feedback signals (issue #255) — 👍/👎 people left on what Last Light wrote,
@@ -30,11 +31,16 @@ import { useTheme } from "../hooks/useTheme";
  */
 
 // Recharts parses fill strings internally and can't resolve `hsl(var(--p))`,
-// so the palettes are literal hex per theme. Same constraint (and the same
-// values) as HomePage's charts.
+// so what remains here is literal hex per theme: `score` is a CATEGORICAL line
+// colour and stays local.
+//
+// `positive` / `negative` are NOT themed and are NOT this page's to pick — a
+// 👍 is the same `good` as a succeeded run, so both come from the shared
+// STATUS palette (issue #329). They used to be per-theme literals here, which
+// drifted from the Home page's and, worse, made the dark pair `#86efac` /
+// `#fca5a5` — ΔE 5.8 for deuteranopes, below the floor, on the two bars in
+// this view that most need telling apart.
 const CHART_DARK = {
-  positive: "#86efac",
-  negative: "#fca5a5",
   score: "#7dd3fc",
   grid: "#21262d",
   axis: "rgba(230, 237, 243, 0.45)",
@@ -43,8 +49,6 @@ const CHART_DARK = {
 };
 
 const CHART_LIGHT = {
-  positive: "#07a06f",
-  negative: "#dc2626",
   score: "#0b3b63",
   grid: "#e2e6ea",
   axis: "rgba(27, 35, 48, 0.55)",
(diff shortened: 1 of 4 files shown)


cliftonc:
@nearform-lastlight can you fix this so we can merge it

cliftonc:
@robinbowes can you fix this one when you get a chance?

Issue #345, opened 2026-08-14
Schema migrations have no formal versioning: failed migrations are silent and downgrades are unguarded


robinbowes:
## Summary

`src/state/migrate.ts` has no formal schema versioning. It is 530 lines of
unordered, self-guarding statements re-executed on every `new StateDb()`, and
the database itself carries no record of which schema it is at. `PRAGMA
user_version` — SQLite's built-in, free, header-resident version integer — is
never read or written anywhere in the codebase.

That is survivable today because of the policy in `spec/10-state.md`:

> Strategy: never drop, never narrow. Long-running deployments accumulate
> schema; SQLite handles it.

The policy is sound. The problem is that **nothing enforces it, and nothing
detects when it has been violated.**

## Three concrete weaknesses

**1. A failed migration is indistinguishable from an applied one.**

All 11 `ALTER TABLE … ADD COLUMN` statements are wrapped like this, and not one
of the 15 `catch` blocks inspects the error:

```ts
try {
  db.exec(`ALTER TABLE workflow_runs ADD COLUMN restart_count INTEGER NOT NULL DEFAULT 0`);
} catch {
  // Column already exists — ignore
}
```

The comment states the intended case, but the code swallows *every* case:
disk full, table missing, a typo in the DDL, a locked database. Boot then
continues with the column absent, and the failure surfaces later as a
`SqliteError` on the first query that touches it — arbitrarily far from the
cause, with nothing in the logs connecting the two.

**2. The downgrade direction is completely unguarded.**

Rolling an image back is the normal recovery move here, and today it is safe
only by convention. New code on an old DB is fine — `migrate()` fixes it
forward. Old code on a *new* DB is fine only because no change has yet dropped
or narrowed anything. Nothing checks. If a change ever does, the rollback path
silently misreads data instead of refusing to start, and there is no
down-migration mechanism to recover with.

**3. Backfills are idempotent by hand-maintained convention.**

`migrate.ts` also carries data backfills, not just DDL — e.g. the
`UPDATE workflow_runs SET owner = …` owner/repo split. These re-run on **every
boot**, forever. The current one is written to be re-runnable, but that is a
property each author has to preserve by inspection. A backfill that is not
perfectly idempotent corrupts a little more on every restart, and the fleet
restarts often.

## What to decide first

Three separable capabilities get conflated here. Worth choosing deliberately
rather than defaulting:

| | buys | cost |
|---|---|---|
[...]

PR #351, opened 2026-08-18
feat(state)!: migrate the state layer to Drizzle ORM — async API, journaled migrations, dual dialect

cliftonc:
Migrates Last Light's state layer from direct `better-sqlite3` to **Drizzle ORM** with an **async store API** and **dual-dialect** support, so a second database is a configuration choice rather than a second rewrite.

Refs #345 — see "What #345 asked for" below; two of its three weaknesses are closed, one is not.

Executable plan, phase docs and every deviation: [`docs/plans/drizzle-migration/`](https://github.com/nearform/lastlight/tree/drizzle-migration/docs/plans/drizzle-migration).

## What changed

- **`better-sqlite3` is gone from the repo.** The driver is `@libsql/client` + `drizzle-orm/libsql` — natively async, so the SQLite and Postgres code paths share one shape (including async transactions), and it ships prebuilt bindings, which is what lets the Dockerfile drop its compiler toolchain.
- **Every store method returns a Promise.** `new StateDb(path)` is replaced by `await StateDb.open(pathOrUrl)` / `StateDb.fromClient(client, dialect)`. `lastlight-workflow-engine`'s `RunStore` / `ExecutionLedger` / `PhaseReporter` ports flip to hard `Promise<T>` (23 engine call sites), so **this is a five-package release**.
- **Migrations are journaled.** `src/state/migrate.ts` — 530 lines of self-guarding DDL re-executed on every boot — is deleted. The schema is declared in `src/state/schema/sqlite.ts`; `drizzle/sqlite/0000_baseline.sql` is a hand-edited fully-idempotent baseline that no-ops over existing production databases, and `0001_backfill_repo_refs.sql` carries the data backfills that used to re-run forever. `__drizzle_migrations` records what has been applied.
- **Postgres runs the whole suite.** `src/state/schema/pg.ts` mirrors all 15 tables with real `jsonb`; the entire state suite and the `SessionManager` suite run a **second time** against real Postgres via PGlite. Portability lives in one seam, `src/state/dialect.ts`. Production stays on sqlite — `StateDb.open()` throws on a `postgres://` URL and no PG driver is a runtime dependency.
- **`database.url` config slot** — `DATABASE_URL` env → overlay → default → `file:` + `dbPath`. Setting none of them is byte-identical to today.
- **Dockerfile drops `python3 make g++` from both stages**; `drizzle/` ships to both the npm tarball and the image via the `files` field.

## Verification

**3,446 tests / 207 files green**, 21/21 turbo tasks. The dual-dialect claim is empirical, not structural: only four tests failed on the first Postgres run.

Three things were checked against real artifacts rather than assumed:

| | Result |
|---|---|
| **Real production data** (41 MB snapshot of drizby, 2,238 executions) | migrations no-op'd in 96 ms; second open 11 ms and idempotent; every row count held; nothing dropped; real `context` / `success` / `extension_status` values round-tripped through the new column mappings |
| **The real image** | builds and boots with **no compiler present at all**, migrates as UID 10001, honours `DATABASE_URL` |
| **A packed npm tarball** | resolves its `migrationsFolder` from an `npm install` and writes a 2-row journal |

## What #345 asked for

| Weakness | Status |
|---|---|
| **1. A failed migration is indistinguishable from an applied one** — 15 `catch` blocks inspecting nothing | **Closed.** No swallowing left; the pre-baseline compat step guards by `PRAGMA table_info` rather than try/catch, so a real failure throws |
| **3. Backfills are idempotent by hand-maintained convention** | **Closed.** They are journaled one-shot migrations. Their statement *order* is load-bearing and documented |
| **2. The downgrade direction is unguarded** | **Not closed.** Drizzle's migrator does not refuse to start when the journal is ahead of the code, and there are no `down` migrations — which #345 itself argues are usually a fiction. The "never drop, never narrow" policy still holds by prose |

Prevention (a destructive-change lint) is likewise not added, though the rule it would enforce is now written down with a concrete cost: production carries two orphan tables nothing declares (`rate_limits`, `system_status`), so `drizzle-kit push` would generate a DROP for both. Generated migrations only.

## Scope and risk

Phases 1–5 of the plan. **Phase 6 — the production Postgres runtime — is deliberately out of this PR**; its doc is included so the reserved slot has a written destination, including a `lastlight server setup` step for the choice.

The rollback story is that the baseline is schema-neutral on current production: every `CREATE` no-ops and the only addition is `__drizzle_migrations`, which the previous image ignores. So the old image reads the same file unchanged.

Remaining before merge: the five-package **v0.26.0** release and the production cutover runbook ([05 §5](https://github.com/nearform/lastlight/blob/drizzle-migration/docs/plans/drizzle-migration/05-config-packaging-release.md)). Re-run the prod-shape smoke first if drizby has moved on.


cliftonc:
All three confirmed and fixed in 7e09603 — thank you, these were real.

`persistPhase`, `appendPhase` and `failWorkflow` all became `Promise<void>` in this PR and none was awaited. The `appendPhase` one is the sharpest: its own comment says *"without this a failed post-review would show as 'pending' despite the run being marked failed"*, and a dropped rejection there produces exactly the state the line exists to prevent. `failWorkflow` was additionally racing the scheduler's own awaited call at scheduler.ts:282, so the run got marked failed by ordering luck.

**Why the review missed it, and what changed as a result.** `post-review.ts` is not in this PR's diff — the signatures changed in `ports.ts`, the breakage is at a caller nothing forced anyone to open. The plan already knew this class of bug (`docs/plans/drizzle-migration/README.md`, "two rules the phases learned the hard way": *`tsc` cannot see a dropped promise*) and its answer was a grep, which is what missed this file.

So the fix is not just the three `await`s. `apps/server/scripts/lint-floating-promises.mjs` now runs from `typecheck`, beside the dep-cruiser boundary gate — `no-floating-promises` implemented against the TS compiler API, since the repo has no eslint setup and one rule doesn't justify adding one. `await` / `void` / `.catch(fn)` / `.then(ok, err)` / `.finally(fn)` are all honoured as deliberate, so intentional fire-and-forget still passes and says so in the source.

Over `apps/server` plus `workflow-engine` and `shared` it reports **exactly these three** and nothing else; zero after the fix. Verified in both directions by reintroducing one (`exit 1` with the bug, `exit 0` without). Two pre-existing `stream.writeSSE` floats in `admin/routes.ts` from April are allow-listed with a written reason — awaiting them is probably correct but changes the behaviour of a live streaming endpoint, which doesn't belong in a state-layer PR.

3,446 tests green with the gate in the `typecheck` task.

PR #353, opened 2026-08-19
chore(deps-dev): batch the stuck Dependabot majors (tailwind 4, daisyUI 5, vite 8)

cliftonc:
Clears **12 of the 21** open Dependabot PRs by doing the majors together in one tree. They had to be batched: they overlap (daisyUI 4 cannot load under Tailwind 4, so those two move as a pair), and each one alone conflicted on `pnpm-lock.yaml`, which is why they were all stuck.

| package | from | to | where |
|---|---|---|---|
| tailwindcss | 3.4.19 | 4.3.3 | both dashboards |
| daisyui | 4.12.24 | 5.7.19 | both dashboards |
| vite | 6.4.3 | 8.2.1 | evals dashboard (server was already 8.1.4) |
| @vitejs/plugin-react | 4.7.0 | 6.0.5 | both dashboards |

## Tailwind 3 → 4 (codemod)

Done with the official `@tailwindcss/upgrade` codemod rather than by hand, because v4 renames utilities that fail *silently* in markup — `rounded` → `rounded-sm`, `min-w-[3rem]` → `min-w-12`. Those generate no error, just wrong pixels, across 11 component files.

Both `tailwind.config.js` files are deleted: content globs are auto-detected in v4 and the font family/size scale is inlined into `@theme`. `postcss.config.js` now uses `@tailwindcss/postcss`, and `autoprefixer` is dropped from both manifests since v4 bundles it.

## daisyUI 4 → 5 (by hand)

No codemod exists for this, and it is where the Dependabot branches would have done damage. #300/#309/#314/#316 carried a bot-authored "make it mergeable" commit that bumped the version and moved the theme block, but **left two silent visual regressions**:

- **Design tokens.** A v5 theme carries more than colours — `--radius-selector`, `--radius-field`, `--radius-box`, `--size-*`, `--border`. Port only the colours and it compiles perfectly clean while every badge, button and table renders square and border-less. (I hit exactly this mid-review; screenshot evidence was a square red `failed` badge.)
- **The light-theme contrast fix.** The muted-text lift read `oklch(var(--bc) / <alpha>)`. v5 has no `--bc` — base-content is a whole colour in `--color-base-content`, not raw oklch components — so all seven rules resolved to nothing and the `neaform` light theme quietly lost the contrast work from `fb26884d`. Now `color-mix(in oklab, ...)`.

Theme colours are otherwise ported value-for-value; no palette change is intended in this PR.

## TypeScript 7 is deliberately excluded

The nine `typescript 5.9.3 -> 7.0.2` PRs (#303/#304/#305/#307/#310/#313/#318/#319/#320) are **not** in this batch. TS 7 moves the entire classic compiler API out of the main entry — `require("typescript")` now exports only `version`, with the API behind `typescript/unstable/*`. That breaks two of this repo's invariant gates:

- **`lint:promises`** hard-crashes (`ts.sys` is undefined). It is a compiler-API consumer by design, and per `apps/server/CLAUDE.md` it exists because a dropped promise is invisible to `tsc` — it caught 14 real bugs during the Drizzle async flip.
- **`lint:boundaries`** is worse, because it *passes*: dependency-cruiser detects TS ≥ 7, warns that it "likely missed TypeScript sources and dependencies", and **exits 0**. The gate becomes hollow while still showing green. Upstream is explicit: "Support for typescript@>=7 will follow when its API is published and stable."

Everything else is ready — with `types: ["node"]` added (TS 7 no longer auto-includes node globals) 11 of 12 packages typecheck clean on TS 7, and `tsc --noEmit` passes for `lastlight-core` too. It is only the two gates that block it.

**Related fix included here:** `packages/workflow-engine` was already on `typescript@^7.0.2` on main, which means its boundary gate has been running degraded. This PR pins it back to `^5.9.3` with the rest, restoring it. That is the one change here that is not a Dependabot bump.

## Verification

`pnpm turbo run typecheck test build` — 21/21 tasks green, 3502 tests passed. Both dashboards build; the compiled CSS was checked to confirm `.badge`/`.btn`/`.table` resolve their radius and border tokens.

Worth a human eye on both dashboards' light (`neaform`) and dark (`lastlight`) themes before merge, since the daisyUI half is a hand port.

## PRs this supersedes

Close on merge: #298 #299 #300 #301 #308 #309 #311 #314 #315 #316 #317 — plus #296, which is already fully superseded by `eb4ac99d` (every target in it is at or below what is on main today).

Leaves open only the nine TypeScript 7 PRs, pending upstream tooling.


cliftonc:
Good catch on `prism-theme.css` — that was a real miss and it's fixed in 21651ce9.

You're right about the mechanism and the blast radius: it's the identical copy in both dashboards, every colour rule used the v4 raw-component variables, and `oklch(var(--bc))` with `--bc` undefined is an invalid colour that browsers drop silently. Code blocks would have rendered as unstyled text on both themes. Applied exactly the mapping you listed — `oklch(var(--x))` → `var(--color-<name>)`, `oklch(var(--x) / α)` → `color-mix(in oklab, var(--color-<name>) α%, transparent)`.

My error was one of method: I grepped `index.css` for `--bc` rather than sweeping every stylesheet, so I confirmed the fix I'd already made instead of looking for the ones I hadn't. I've now verified against the **built** CSS rather than the source — no `oklch(var(--xx))` survives in either bundle, and `.token.comment` / `.token.keyword` resolve to real colours.

Since one leftover implies there may be others, I swept the rest of the v5 breaking-change surface. Three findings, no further fixes needed:

- `diff-theme.css` (evals) uses hardcoded hex, not daisyUI variables — unaffected. Its `.ll-diff .diff` specificity override still holds, since v5 keeps its own `.diff` component.
- Nothing reads daisyUI variables from JS (no `getComputedStyle`/`getPropertyValue`), so there's no runtime path with the same problem.
- `input-bordered` / `select-bordered` / `textarea-bordered` (18 uses) are **inert, not broken**. v5 drops them because `.input` now carries `border: var(--border) solid var(--input-color)` by default — I confirmed that rule in the built CSS, and `--border` is set by the design tokens in the theme blocks. Rendering is unchanged, so they're dead markup. I've deliberately left them rather than widen this diff; happy to strip them here or in a follow-up if you'd prefer.

Also fixed a stale pointer the migration created: both files' header comments referenced `tailwind.config.js`, which this PR deletes. They now point at the `@plugin "daisyui/theme"` blocks in `index.css` and spell out the v4→v5 variable trap, since the whole point of that file is that theme edits flow through it automatically.

Gate re-run after the fix: `pnpm turbo run typecheck test build` — 21/21 tasks, 3502 tests.

Issue #205, opened 2026-07-21
User identity records + actor logging (foundation for team visibility)


cliftonc:
## Problem

Every login path resolves a real identity — GitHub `login` (`src/admin/routes.ts` ~737), Slack `sub`/`user_id`/`team_id` (`routes.ts` ~618) — then collapses it to a single `method` string in the HMAC token (`TokenPayload = { exp, method? }`, `src/admin/auth.ts`). There is **no `users` table** anywhere. "Who did it" survives only as free-text audit columns (`workflow_approvals.requested_by`/`responded_by`, `cron_overrides.updated_by`) or as rendered prompt text (`Requested by: X` in `src/workflows/simple.ts`), and dashboard/CLI actions are hardcoded to `"admin"` / `"cli"`. We can't attribute a workflow run to a person, show their avatar/name, or email them.

This was split out of #169 as its **foundation**: establish a real user identity + actor logging here; team-based per-repo visibility (#169) then builds on it.

## Goals

- A first-class `users` record capturing GitHub identity (numeric id, login, name, **email**, **avatar_url**) plus a linked Slack user id — so the dashboard can show real people and we can email them later.
- Populate/refresh it on every dashboard login (GitHub + Slack OAuth callbacks).
- Carry the verified `login` in the session token (not just `method`).
- **Actor logging**: persist *who* triggered/acted on a workflow across every invocation path — CLI build/run, manual cron fire, retry, cancel, approval respond, GitHub webhook, Slack.
- **Slack → user matching**: resolve a Slack user's email and match it to a `users` row so a Slack-initiated run/approval attributes to the same person.

## Non-goals

- No teams / per-repo visibility — that's #169, which builds on this.
- No outbound email plumbing yet — just **store** emails as the future hook (the repo has zero email capability today; a sender is a separate later issue, e.g. via the `cloudflare-email-service` skill).
- No FK/normalization of the existing free-text actor columns; `users` is an additive enrichment table joined on `login`.

### 1. `users` table + `UserStore`

Adapt guidemode's arctic-based Postgres pattern to Last Light's SQLite / raw-SQL. Add to the top `db.exec` block in `src/state/migrate.ts`:

```sql
CREATE TABLE IF NOT EXISTS users (
  id TEXT PRIMARY KEY,                    -- randomUUID
  github_id INTEGER UNIQUE,               -- stable numeric id (upsert key); null for Slack-only rows
  login TEXT UNIQUE,                      -- GitHub login = the soft join key used everywhere
[...]

PR #364, opened 2026-08-27
Activity log: one append-only stream of who did what (#206)

### LINKED ISSUE #206: Activity Log: single append-only audit stream of all user actions
## Problem

#205 established a real `users` identity + per-run/per-execution actor attribution, but attribution is **scattered across ledgers**: a run's origin lives on `workflow_runs.triggered_by`, retry/cancel/approve actors on the `executions` ledger and `workflow_approvals.responded_by`, and config toggles on `cron_overrides.updated_by` / `workflow_overrides.updated_by`. There is **no single chronological stream** of "who did what, when" — so you can't answer "what has this user done across the system?" or "what happened on this instance in the last hour?" without joining five tables, and many actions (login, config edits, kill-switch flips, CLI triggers) leave no durable trace at all.

This is the follow-up #205 explicitly deferred: a first-class **Activity Log** — one append-only row per meaningful user/system action.

## Goal

A single `activity_log` table (append-only, one row per action) capturing **actor + verb + target + when + outcome**, written from every action surface (dashboard API, CLI/API, Slack, GitHub-initiated, cron, system). Surfaced read-only in the dashboard (a global feed + per-user and per-run filters) and queryable via the admin API / `lastlight` CLI.

It **complements** #205's columns (which stay as the fast per-run/per-execution attribution); the Activity Log is the cross-cutting audit stream layered on top, joined to `users` on `login`.

## High-level shape

`activity_log` (append-only; never updated/deleted):

- `id`, `created_at`
- `actor_login` (free-text, joins `users.login`) + `actor_type` (`github | slack | cli | cron | admin | system`) — reuse #205's `TriggerActorType`
- `action` — a stable verb enum, e.g. `login`, `workflow.trigger`, `workflow.retry`, `workflow.cancel`, `approval.approve`, `approval.reject`, `cron.toggle`, `cron.trigger`, `workflow.toggle`, `config.edit`
- `target_type` + `target_id` — e.g. `workflow_run:<id>`, `cron:<name>`, `repo:<owner/repo>`, `approval:<id>`
- `outcome` (`ok | denied | error`) + optional `detail` (short string / small JSON)
- indexes on `(created_at)`, `(actor_login, created_at)`, `(target_type, target_id)`

### Write seam

One thin `ActivityStore.record({...})` (mirrors the #205 stores) called from a **single choke point per surface** — ideally a small helper the actor-hardcoded admin routes already funnel through (they now all read `actorFromContext(c)` after #205), plus `dispatchWorkflow` for triggers and the Slack/GitHub connectors for external actions. Logging is **best-effort and never blocks the action** (swallow store errors, like #205's identity capture).

### Read surface

- `GET /admin/api/activity` — paginated feed with `actor` / `action` / `target` / `since` filters + a companion count (mirror the `workflow-runs` list shape).
- Dashboard: a global "Activity" tab + an inline per-run activity strip on the run detail page + a per-user view once #169's user pages exist.
- `lastlight activity` CLI subcommand (thin admin-API client).

## Non-goals

- Not replacing #205's per-run/per-execution actor columns — those stay as the hot-path attribution; this is the audit stream on top.
- No retention/rotation policy in v1 (append-only; a later issue can add pruning if volume warrants).
- No PII beyond what login/email already capture; `detail` should stay a short summary, not full payloads.

## Acceptance criteria

- Every user-initiated action across dashboard / CLI / Slack / GitHub / cron writes exactly one `activity_log` row with actor + action + target + outcome.
- The write is best-effort — a store failure never fails the underlying action.
- `GET /admin/api/activity` returns a filterable, paginated feed; the dashboard renders a global feed + per-run strip.
- Rows join to `users` on `actor_login` for name/avatar enrichment (reusing #205's identity table).
- Covered by Vitest following the existing `tests/state/` + `tests/admin/` patterns.

## Depends on

- #205 (user identity + actor attribution) — this builds directly on its `users` table, `TriggerActorType`, and `actorFromContext` seam.

lv10:
Closes #206 — in progress. **Draft: Phase 1 of 3 is complete**; Phases 2 and 3 land as further commits on this branch.

## What this is

`activity_log` — one append-only row per **user-initiated** action, carrying actor + verb + target + outcome, across the dashboard, CLI, Slack, GitHub and cron.

Today that question has no answer: #205 put a real actor on every run and execution, but attribution is spread across five ledgers (`workflow_runs.triggered_by`, `executions`, `workflow_approvals.responded_by`, `cron_overrides.updated_by`, `workflow_overrides.updated_by`), and login, config edits, container kills and artifact edits write to none of them.

This **complements** #205 rather than replacing it. `triggered_by` stays the hot-path per-run attribution; this is the audit stream layered on top, soft-joined to `users` on `login`.

## Read this first

**[`docs/plans/activity-log/`](docs/plans/activity-log/)** — the design, the locked decisions, and one phase doc per commit group. Each phase doc gains Execution notes as it lands, so the directory reads as both plan and record.

Three decisions were reached **against** the issue's own instructions, and are worth a look before the code:

1. **The issue body is stale.** It predates #351/#352 and says to add DDL to `src/state/migrate.ts`, which no longer exists in that form. Everything schema-shaped here follows `src/state/CLAUDE.md` instead.
2. **No middleware seam exists** to hang the write on — no `onError`, no `notFound`, no shared response helper. So: explicit calls at each site, with a table-driven test pinning the route→verb map in place of the coverage a middleware would have bought.
3. **System fan-out is recorded once at its cause, not per dispatch.** A cron fan-out dispatches per repo, so logging each as an action would make the dominant row source a thing no human did — the same confusion `cron_runs` avoids by keying on the cron rather than the workflow. That bounds growth *below* `workflow_runs`, which is what makes deferring retention safe.

## Phase 1 — schema + store (this draft)

- `activity_log` on both dialects; `detail` diverges `text({mode:"json"})` → `jsonb` with a shared `$type<T>()`.
- Both migrations generated and committed unedited, same commit. Purely additive: one `CREATE TABLE`, three `CREATE INDEX`, no `DROP`.
- `ActivityStore` (`record` / `list` / `actions`) on the `CronRunStore` pattern, tables via `tablesOf(client)`. Ids creation-ordered; both reads tie-break on `id`, because a page boundary inside a same-millisecond run would otherwise skip or repeat rows.
- Behavioural tests in `tests/state/suites/`, so both the sqlite and PGlite legs replay identical bodies.

## Two things a reviewer should look at closely

**`schema-equivalence.test.ts` needed restructuring, not a bumped count.** It asserted the schema is *identical* before and after boot — sound while every table came from the baseline, but `activity_log` is the first table a **post-baseline** migration creates, so the two sets legitimately differ. Rather than weaken it, `LEGACY_TABLES` (the pre-Drizzle production shape the baseline must still no-op over) is now separate from a derived `EXPECTED_TABLES`, with `POST_BASELINE_TABLES` / `_INDEXES` naming the difference. The proof the file exists for is intact.

**`src/state/CLAUDE.md`'s add-a-table checklist said three places; there are six.** The three it named fail loudly; the other three are hardcoded counts in the test suite that fail as an off-by-one reading like a bug in your own change. The checklist is corrected in this PR.

## Verification

`pnpm turbo run typecheck test build` — 25/25 tasks, 3845 tests, on `main` at `2e49dddd`. State suite green on **both** dialects; `lint:promises` and `lint:boundaries` clean. `pg-client.ts` is untouched, so the `PG_INTEGRATION=1` leg is not implicated.

## Still to come

- **Phase 2** — `recordActivity()` and its 19 call sites, plus the three routes still hardcoding `"admin"` as the actor (`routes.ts:1868`, `:2511`, `:2556`).
- **Phase 3** — `GET /admin/api/activity`, the dashboard tab + per-run strip, `lastlight activity`.

## Open questions

Both cheap now, expensive once rows exist:

1. **The verb vocabulary** — fourteen verbs, a closed union in `activity-store.ts`.
2. **`config.edit` does double duty** for setting and clearing a cron schedule override, distinguished only by `detail`. The alternative is `cron.schedule` + `cron.schedule.clear`. I lean toward keeping it broad since #206 names it as one verb and the target already carries the `cron:<name>` specificity.

### DIFF
--- a/apps/server/CLAUDE.md
+++ b/apps/server/CLAUDE.md
@@ -323,7 +323,7 @@ src/
   state/
     db.ts               `StateDb` — the ASYNC factory (`await StateDb.open(url)`
                         / `StateDb.fromClient(client, dialect)`; no public
-                        constructor) that wires the seven stores together and
+                        constructor) that wires the eight stores together and
                         is the single import surface for their types. Every
                         store method returns a Promise. `open()` picks the
                         engine off the URL: libsql for `file:`/`:memory:`/a
@@ -347,10 +347,10 @@ src/
                         ships in the agent image — what `lastlight server db`
                         runs inside the container, since the CLI may never gain
                         an edge to core.
-    schema/sqlite.ts    Drizzle schema — the source of truth for all fifteen
+    schema/sqlite.ts    Drizzle schema — the source of truth for all sixteen
                         tables (executions, workflow_runs, workflow_approvals,
                         cron_runs, cron_overrides, workflow_overrides, users,
-                        messaging_*, feedback_*, github_team*).
+                        activity_log, messaging_*, feedback_*, github_team*).
     schema/pg.ts        The name-parity pgTable mirror. NOTHING under src/ may
                         import it — it exists for drizzle-kit and the PGlite
                         test leg. A schema change means editing BOTH files and
@@ -377,6 +377,17 @@ src/
                         only record that it ran at all. Keyed on the cron
                         rather than the workflow so a run dispatched by
                         `/api/run` or a comment cannot skew a cron's health.
+    activity-store.ts   The `activity_log` audit stream (issue #206) — one
+                        append-only row per USER-INITIATED action, across the
+                        dashboard, CLI, Slack, GitHub and cron. Complements
+                        #205's per-run `triggered_by` columns rather than
+                        replacing them: those stay the hot-path attribution, and
+                        this is the chronological stream that answers "what has
+                        this person done?" without joining five ledgers. System
+                        fan-out is deliberately NOT recorded per dispatch — a
+                        cron fires once and is logged once (`cron.fire`), the
+                        same reason `cron_runs` keys on the cron rather than the
+                        workflow.
     team-store.ts       The dashboard's per-repo visibility CACHE (issue #169):
                         github_teams / _team_repos / _team_members /
                         github_visibility_sync. Not a mirror of the org — rows
--- a/apps/server/dashboard/src/App.tsx
+++ b/apps/server/dashboard/src/App.tsx
@@ -29,6 +29,7 @@ const RouterPlayground = lazy(() =>
   import("./components/RouterPlayground").then((m) => ({ default: m.RouterPlayground })),
 );
 import { FeedbackPage } from "./components/FeedbackPage";
+import { ActivityPage } from "./components/ActivityPage";
 import {
   HomeIcon,
   PlayCircleIcon,
@@ -41,6 +42,7 @@ import {
   CommandLineIcon,
   ShareIcon,
   HandThumbUpIcon,
+  ListBulletIcon,
 } from "@heroicons/react/24/outline";
 import {
   useUrlState,
@@ -53,11 +55,11 @@ import {
 } from "./hooks/useUrlState";
 
 type AuthState = "checking" | "required" | "ok";
-type Tab = "home" | "sessions" | "chat-sessions" | "workflows" | "runs" | "repos" | "crons" | "feedback" | "logs" | "config" | "router-playground";
+type Tab = "home" | "sessions" | "chat-sessions" | "workflows" | "runs" | "repos" | "crons" | "feedback" | "logs" | "config" | "router-playground" | "activity";
 
 const PAGE_SIZE = 50;
 
-const TABS = ["home", "workflows", "runs", "sessions", "chat-sessions", "repos", "crons", "feedback", "logs", "config", "router-playground"] as const;
+const TABS = ["home", "workflows", "runs", "sessions", "chat-sessions", "repos", "crons", "feedback", "activity", "logs", "config", "router-playground"] as const;
 
 const SESSION_SOURCE_PATHS: Record<"sessions" | "chat-sessions", string> = {
   sessions: "/admin/api/sessions",
@@ -305,6 +307,7 @@ function Dashboard({ onLogout }: { onLogout: () => void }) {
               { id: "repos", label: "Repos", Icon: FolderIcon },
               { id: "crons", label: "Crons", Icon: ClockIcon },
               { id: "feedback", label: "Feedback", Icon: HandThumbUpIcon },
+              { id: "activity", label: "Activity", Icon: ListBulletIcon },
               { id: "logs", label: "Logs", Icon: CommandLineIcon },
               { id: "router-playground", label: "Router Playground", Icon: ShareIcon },
               { id: "config", label: "Config", Icon: Cog6ToothIcon },
@@ -402,6 +405,8 @@ function Dashboard({ onLogout }: { onLogout: () => void }) {
         <LogsPage />
       ) : tab === "feedback" ? (
         <FeedbackPage />
+      ) : tab === "activity" ? (
+        <ActivityPage />
       ) : tab === "config" ? (
         <ConfigPage />
       ) : tab === "repos" ? (
--- a/apps/server/dashboard/src/api.ts
+++ b/apps/server/dashboard/src/api.ts
@@ -571,6 +571,34 @@ export interface DailyStat extends OutcomeCounts {
 // copied here by hand.
 
 /** One 👍/👎 on something the bot wrote, scored against the run that wrote it. */
+/**
+ * One row of the audit stream (issue #206).
+ *
+ * HAND-MIRRORED from `src/state/activity-store.ts` — the dashboard has no
+ * import edge to core. `tests/admin/dashboard-activity-mirror.test.ts` pins the
+ * two together, because a mirror like this drifted once before and hid three
+ * config blocks for a release.
+ *
+ * `actorLogin` is optional on purpose: a password session and an auth-disabled
+ * instance both write a row with no verified login. Render that as "no login",
+ * never as an empty cell — the row is not missing data, the login genuinely
+ * does not exist.
+ */
+export interface ActivityRecord {
+  id: string;
+  createdAt: string;
+  actorLogin?: string;
+  actorType?: "github" | "slack" | "cli" | "cron" | "admin" | "system";
+  action: string;
+  targetType?: string;
+  targetId?: string;
+  outcome: "ok" | "denied" | "error";
+  detail?: Record<string, string | number | boolean>;
+}
+
+/** `users.login` → identity, for avatar + real name. Absent when unresolved. */
+export type ActivityUsers = Record<string, { login?: string; name?: string; avatarUrl?: string }>;
+
 export interface FeedbackSignal {
   id: string;
   anchorId: string;
@@ -856,6 +884,31 @@ export const api = {
   dailyStats: (days = 30) => req<{ daily: DailyStat[] }>(`/stats/daily?days=${days}`),
   hourlyStats: (hours = 24) =>
     req<{ hourly: DailyStat[] }>(`/stats/hourly?hours=${hours}`),
+  activity: (
+    opts: {
+      limit?: number;
+      offset?: number;
+      actor?: string;
+      action?: string;
+      /** `<type>:<id>` — e.g. `workflow_run:4f3a…`. Also serves the per-run strip. */
+      target?: string;
+      since?: string;
+    } = {},
+  ) => {
+    const qs = new URLSearchParams();
+    if (opts.limit) qs.set("limit", String(opts.limit));
+    if (opts.offset) qs.set("offset", String(opts.offset));
+    if (opts.actor) qs.set("actor", opts.actor);
+    if (opts.action) qs.set("action", opts.action);
+    if (opts.target) qs.set("target", opts.target);
+    if (opts.since) qs.set("since", opts.since);
+    const q = qs.toString();
+    return req<{ activity: ActivityRecord[]; total: number; users: ActivityUsers }>(
+      `/activity${q ? `?${q}` : ""}`,
+    );
+  },
+  activityActions: () => req<{ actions: string[] }>("/activity/actions"),
+
   feedbackSignals: (opts: { limit?: number; workflow?: string; source?: string } = {}) => {
     const qs = new URLSearchParams();
     if (opts.limit) qs.set("limit", String(opts.limit));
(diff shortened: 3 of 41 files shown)


cliftonc:
Looks good to me - the schema equivalence test I suspect we can do in a much smarter way to ensure it's ensuring sqlite and postgres stay in sync without this overhead, but otherwise this looks fine to continue.

lv10:
@cliftonc — deliberately still a draft (Phases 2 and 3 land on this branch), but **two decisions would be much cheaper to change now than later**, and I'd rather not build 19 call sites on top of the wrong answer. No rush on the rest of the review.

Both are in [`docs/plans/activity-log/README.md`](docs/plans/activity-log/README.md) → *Open questions*.

### 1. The verb vocabulary

Fourteen verbs, `<noun>.<verb>`, declared as a closed union in [`src/state/activity-store.ts`](../blob/feat/activity-log/apps/server/src/state/activity-store.ts):

```
login
workflow.trigger  workflow.retry  workflow.cancel  workflow.toggle
approval.approve  approval.reject
cron.fire  cron.trigger  cron.toggle
config.edit
container.kill  artifact.edit  pr.retry
```

Right now this is a one-file edit. Phase 2 hardcodes them across 19 call sites; after that it's data.

### 2. `config.edit` does double duty

It covers both setting a cron schedule override and clearing it, distinguished only by `detail` (`{ schedule }` vs `{ cleared: true }`). The alternative is `cron.schedule` + `cron.schedule.clear`.

I lean toward keeping it broad — #206 names `config.edit` as one verb, and `target_id` already carries the `cron:<name>` specificity — but it's the one verb whose granularity I'm genuinely unsure about.

### Also worth a sanity check: the volume decision

Not a question so much as the thing most likely to be wrong. `workflow.trigger` is written **only for a human actor type** (`github`/`slack`/`cli`/`admin`), so a cron fan-out's per-repo dispatches are *not* logged as actions — the fire is logged once, at its cause, as `cron.fire`. That is #206's own "user-initiated" wording, and the same reasoning as `cron_runs` keying on the cron rather than the workflow (`spec/10-state.md` L406).

The consequence is that `activity_log` grows more slowly than `workflow_runs` already does — roughly 150 cron rows/day on the current schedules, plus a strict subset of run creations — which is what makes #206's deferral of retention safe rather than optimistic. Reasoning is in [README.md → "Decision 3, and the volume question it answers"](docs/plans/activity-log/README.md).

If you'd rather every dispatch were recorded, that's fine too — but it changes the retention story, so better to know before Phase 2.

### One thing this turned up that outlives the feature

`src/state/CLAUDE.md`'s *"If you added a table"* said three places don't update themselves. Adding one found **six** — the other three are hardcoded counts in the test suite that fail as an off-by-one reading like a bug in your own change. Corrected in the last commit.

And `schema-equivalence.test.ts` needed restructuring rather than a bumped count: it asserted the schema is *identical* before and after boot, which was sound while every table came from the baseline. `activity_log` is the first table a **post-baseline** migration creates, so `before` and `after` now legitimately differ. Rather than weaken it, `LEGACY_TABLES` is split from a derived `EXPECTED_TABLES`. Worth a look — it's the one change in here that touches the proof that file exists for.

cliftonc:
1. This is fine by me, they are simple strings that are what they say they are.

2. Config.edit is fine for now, there is very little today that is editable.

lv10:
> the schema equivalence test I suspect we can do in a much smarter way to ensure it's ensuring sqlite and postgres stay in sync without this overhead

Done — pushed in 9ecfde22. One clarification first, because it changed what the fix was:

**Those are two different files.** `schema-parity.test.ts` is the sqlite↔pg sync test, and it was already derived from the schema exports. `schema-equivalence.test.ts` is **sqlite-only** — it proves the Drizzle *baseline* no-ops over the legacy pre-Drizzle production shape rebuilt from `fixtures/legacy-schema.sql`. The bookkeeping you flagged was mine, and it was in the second one, which isn't about Postgres at all.

### Why the overhead existed

That test applied **every** migration and then subtracted back out whatever later migrations were allowed to add. `activity_log` is the first table a post-baseline migration creates, so it was the first time `before` and `after` legitimately differed — and my fix was to hand-list the difference. That's the overhead: it grows by one entry per table, forever.

### What it does now

Split the run, so the test asserts its actual invariant:

1. Apply the **baseline alone** → assert nothing changed. No exceptions list, because at that point there genuinely are no exceptions.
2. Apply the **full set** → assert the result matches the tables `schema/sqlite.ts` *declares*.

The second is stronger than the list it replaced: the migrations and the schema module now have to agree about what exists. `LEGACY_TABLES` is read off the fixture text, so both lists are derived and neither needs touching again.

**Verified non-vacuous by mutation**: pointing the baseline-only run at all three migrations makes the test fail. The split is load-bearing, not cosmetic.

### Three counts also retired

Same reasoning, wherever the assertion above already proved the thing:

- `schema-parity`'s `covers all N tables` — the two set-difference checks already prove sync.
- `data-migrate`'s two counts — `assertCoversEveryTable()` already checks **both** directions (missing from `TABLE_ORDER`, and stale in it).

Each only additionally caught a table removed from two places at once — a deliberate act — at the price of an edit on every table added.

**Net effect**: `src/state/CLAUDE.md`'s add-a-table checklist goes from the six entries I'd expanded it to, back down to **three** — what it said before this PR touched it. The only remaining manual bump is `MIGRATION_COUNT`, which is deliberately not derived from the journal per its own comment. I left that alone.

---

### Not done, deliberately — two things I'd rather you decide

**1. Generating `pg.ts` from `sqlite.ts`.** If the underlying worry is "keep sqlite and postgres in sync", the structural answer is to stop hand-mirroring: codegen with an explicit mapping for the four documented divergences (`integer+boolean→boolean`, `text+json→jsonb`, `real→doublePrecision`, `autoincrement→identity`) would make drift *impossible* rather than merely tested. It also gives up the hand-written per-dialect comments in `pg.ts`, so it's a real trade and a separate issue, not a drive-by in a feature PR.

**2. `scripts/lint-floating-promises.mjs`'s `ALLOWED` set is keyed by `file:line`.** This branch's two added imports shifted the two deliberately-unfixed `stream.writeSSE` entries from 452/458 to 454/460, and the gate failed pointing at SSE code this PR never touched. I updated the numbers and added a comment about the coupling, but did **not** re-key it on expression text — that file says to fix its contents in a change that's about SSE, and redesigning a gate as a drive-by seemed like the same mistake pointed the other way. Happy to do it if you'd rather.

Phase 2 (the write seams) is also on this branch now and green. Phase 3 — the read surfaces — is next. The verb-vocabulary question in my earlier comment is still open, and those verbs are now referenced across 22 call sites, so it's costlier to change than it was — though nothing is deployed and no rows exist yet.

lv10:
Thanks — the first finding was a real bug and I'd missed it. All three phases are now on this branch and CI is green (`check` 6m32s, `pg-integration` 1m8s on `1bd5179c`).

## 1. `cron.fire` actor type — fixed in c3977269

Correct diagnosis, and the reasoning was exactly right: the trigger route put `_cronActor` (the login) on the context but nothing about **how** they authenticated, so both fire paths could only guess — and guessed `"admin"`. A GitHub-authenticated "Run now" therefore wrote `cron.trigger` as `github` and its paired `cron.fire` as `admin`: two rows for one action disagreeing about the same person.

Threads `_cronActorType` alongside `_cronActor`, read in both `runner.ts` and `handlers.ts`. `admin` survives only as the genuine-unknown fallback — which is what it actually means, a password session. Membership-checked through `isTriggerActorType` rather than cast, the same rule `dispatchWorkflow` applies to the untrusted dispatch context.

## 2. No `GET /admin/api/activity` — that was Phase 3, now landed

Right that it was missing; it was the next phase rather than an oversight, and your suggestion to note the deferral on `list()` was the correct call for the state the branch was in when you read it. It's moot now — the endpoint exists (d8cfeece), along with the dashboard tab, a per-run strip, and `lastlight activity`.

Two things there worth a look, since they were judgement calls rather than mechanics:

- **`?target=` splits on the FIRST colon.** A target id can contain one (`container:lastlight-sandbox-a:b`); the type never does. A plain `split(":")` would silently truncate the id.
- **It is deliberately not repo-scoped.** `?repos=` on `/workflow-runs` and `/sessions` is UI declutter rather than authorization — those endpoints keep returning global data and the filtering is client-side. An audit stream silently narrowed by which teams you happen to belong to would mislead in a way a run list does not. Narrowing this one is a real authorization decision, so I left it out and said so in the code. Happy to add it if you read that differently.

## 3. `"test-secret"` across 14 files — not in this PR, and I'd argue not as stated

Real observation, and I'm not going to pretend the duplication is good. But two reasons to leave it:

**It's pre-existing in 13 files this branch never touches.** Editing them here mixes a test-hygiene refactor into a feature PR, which is the same drive-by problem I avoided with the SSE allowlist below.

**The stated benefit doesn't quite hold.** "A future change is a single-file edit" assumes somebody will change it — but nobody ever changes a test secret; it's an arbitrary string that exists to be non-empty. The genuine benefit is smaller and different: consistency, and one obvious place to look when writing test number 15.

So: worth doing, worth doing separately. I'll file it as a chore issue unless you'd rather I fold it in.

---

## Two things I deliberately left for you

Neither is mine to decide unilaterally, and both are recorded in the phase docs:

**Generating `pg.ts` from `sqlite.ts`.** If the underlying worry behind your schema-test note is "keep the dialects in sync", the structural answer is to stop hand-mirroring — codegen with an explicit mapping for the four documented divergences would make drift impossible rather than merely tested. It gives up the hand-written per-dialect comments in `pg.ts`, so it's a real trade and a separate issue.

**`scripts/lint-floating-promises.mjs`'s `ALLOWED` set is keyed by `file:line`.** Two added imports in this branch shifted the two deliberately-unfixed `stream.writeSSE` entries from 452/458 to 454/460, and the gate failed pointing at SSE code this PR never touched. I updated the numbers and added a comment about the coupling, but didn't re-key it on expression text — that file says to fix its contents in a change that's about SSE, and redesigning a gate as a drive-by seemed like the same mistake pointed the other way.

## Note on this review's freshness

`last-light/review` has come back `skipping` on the last two pushes (18m each), so this review reflects the branch as of Phase 2 and hasn't re-run against Phase 3. Findings 1 and 2 are addressed regardless, but if you want a fresh pass it'll need a nudge.

cliftonc:
Thanks @lv10!

PR #374, opened 2026-09-08
feat(config): point providers at your own LLM gateway (#373)

### LINKED ISSUE #373: Model endpoints are hardcoded: no way to point at a self-hosted LLM gateway
## Report from a user

> It also had hardcoded URLs for the AI models, which I changed in my local to use my local gateway.

They patched the source to point Last Light at their own LLM gateway. That should be configuration.

## Where we actually stand

Confirmed. `packages/shared/src/providers.ts` is the single source of truth for providers, and `baseUrl` is a compile-time constant per entry (`https://api.anthropic.com/v1`, `https://api.openai.com/v1`, and ~15 more). Nothing reads an override:

- `apps/server/src/engine/llm.ts` (the cheap one-shot helper used by the screener + intent classifier) builds its URL straight from `spec.baseUrl` — `anthropicUrl()` / `openaiUrl()`. The bare-model-id fallback even synthesises a pseudo-spec hardcoded to `https://api.openai.com/v1`.
- There is no `ANTHROPIC_BASE_URL` / `OPENAI_BASE_URL` / `models.*.baseUrl` support anywhere in the tree; the only base-URL override that exists is for the **GitHub** API (`config.githubApiBaseUrl` / `GITHUB_API_URL`, in agentic-pi), described there as a test/eval escape hatch.
- The sandbox **egress allowlist** is derived from the same registry (`PROVIDER_HOSTS` → `apps/server/src/sandbox/egress-allowlist.ts`). So even with an overridable URL, a gateway on an unlisted host is blocked unless the phase sets `unrestricted_egress: true`. Both halves have to move together.
- Partial escape hatch that exists but is unwired: agentic-pi's `resolveModel` falls back to pi's `ModelRegistry` for custom models (`models.json`, loaded by `ModelRuntime`), so a custom provider/endpoint *could* be registered there — but nothing in `apps/server` ever provisions that file, so it isn't reachable from a Last Light deployment.

Worth naming the reasons this matters beyond one user's convenience: a gateway is how a deployment gets central spend accounting, key custody, rate limiting and audit — and self-hosted or Azure/Bedrock-fronted models are simply unreachable today unless they happen to be one of the registry's 18 entries.

## What this issue is asking for

Let a deployment point the model calls at an arbitrary OpenAI-compatible or Anthropic-compatible endpoint, without patching source.

Sketch:

1. **Per-provider `baseUrl` override.** Config (`models:` / provider block) plus an env layer, resolved on top of the registry default so unset deployments behave byte-for-byte as today. The `api` family (`openai-completions` / `anthropic-messages`) stays the thing that picks the request shape, which is what makes most gateways just work.
2. **Custom provider entries.** A deployment declares `{ prefix, baseUrl, api, envKey, host }` for a gateway that isn't in the registry at all, rather than being forced to masquerade as `openai`.
3. **Egress:** feed the resolved hosts into the allowlist automatically — the same way `OTEL_EXPORTER_OTLP_*_ENDPOINT` hosts are already parsed and normalised into it (`parseCollectorHosts` in `apps/server/src/config/config.ts`). That precedent is exactly the pattern to copy.
4. **The sandbox path too.** Core forwards provider API-key env vars into the sandbox (`PROVIDER_ENV_KEYS`, `apps/server/src/engine/agent-executor.ts`); the base-URL override has to reach the in-sandbox model call as well, not just the in-process `llm.ts` helper. Registering a custom model in pi's `models.json` is the likely mechanism.

Notes:

- Keep the wizard (`packages/cli/src/setup.ts`) honest — a custom endpoint should be selectable there or at least documented, since the wizard is where most deployments meet the provider list.
- A gateway URL is not a secret but often carries an internal hostname; it belongs in config, and the key stays in `secrets/.env` as today.
- Should any URL be allowed, or must it be https? Suggest https-only by default with an explicit opt-out for `http://localhost`-style gateways, since that is precisely the reported case.

cliftonc:
Closes #373.

Model endpoints were compile-time constants in `packages/shared/src/providers.ts`, so pointing Last Light at a self-hosted or corporate LLM gateway meant patching source — which is exactly what the reporting user did. This makes it configuration.

## What you can now write

```yaml
providers:
  # A gateway speaking Anthropic's dialect: only the URL moves. Models, request
  # shape and ANTHROPIC_API_KEY are inherited from the registry entry.
  anthropic:
    baseUrl: https://gateway.internal/anthropic

  # A provider the registry has never heard of. baseUrl is required (nothing to
  # inherit); api defaults to openai-completions and envKey to ACME_API_KEY.
  acme:
    baseUrl: https://llm.corp.example/v1
    api: openai-completions
    envKey: ACME_API_KEY
```

Env equivalents: `ANTHROPIC_BASE_URL` / `OPENAI_BASE_URL` / … (derived from the prefix, so they match the names the vendor SDKs already use) and `LASTLIGHT_PROVIDERS` (JSON — the only env route that can declare a custom provider). Both merge **per prefix** with the config file, so moving one endpoint from the env doesn't drop a custom provider declared in `config.yaml`.

## How it reaches all four model paths

`resolveProviderRegistry()` (in `lastlight-shared`, pure and total — no overrides resolves byte-for-byte to today's registry) applies the block over the shipped defaults. `loadConfig()` resolves it **once** and installs it in the new `apps/server/src/config/provider-registry.ts`; every consumer reads the resolved registry instead of the shipped constants:

| Path | What follows the override |
|---|---|
| `src/engine/llm.ts` (screener + classifier) | request URL, api family, key env var — **including the bare-model-id fallback**, which synthesized a pseudo-spec hardcoded to `https://api.openai.com/v1` |
| the sandbox (agentic-pi) | `AGENTIC_PI_PROVIDERS` in the sandbox env (docker / smol / k8s, where the model call is in-guest) and the new `providers` run option (gondolin / none). agentic-pi registers the override through pi's own `ModelRuntime.registerProvider` seam, so a known provider keeps its models, auth and cost table |
| the egress allowlist | `DEFAULT_ALLOWLIST` became `defaultAllowlist()` and `PROVIDER_HOSTS` became `providerHosts()`, reading the resolved registry — so the firewall permits the gateway instead of denying it. Same automation as the OTEL collector hosts, one layer down |
| in-process chat (`chat-runner.ts`) | patched `baseUrl`, or a model synthesized from the override for a provider pi-ai has no catalog entry for. Not in the issue's list, but leaving it out would mean chat alone kept talking to the vendor |

The registry lives in `shared` for the usual reason: core needs it at runtime and the CLI needs it offline — for the wizard's new **"Self-hosted / gateway endpoint"** provider option, which validates the URL with the same function the server applies at boot and writes the `providers:` block into the overlay.

## Decisions worth reviewing

- **A bad override fails the boot**, rather than warning and continuing. Continuing would send the operator's prompts and API key to the vendor — the precise failure the feature exists to prevent.
- **https required, loopback excepted.** `http://localhost:4000` needs no opt-in (it's the reported case, and there's no network to eavesdrop on); `http://` anywhere else needs `LASTLIGHT_ALLOW_INSECURE_PROVIDER_URLS=1`.
- **The URL goes in the committed overlay, the key does not.** A gateway URL is deployment routing — the mirror image of the `database.url` rule, and the setup test pins both directions.
- **Custom providers register zero cost rates.** A gateway publishes no price list, so a run against one reports *no* spend rather than a wrong one. Worth a look if anyone would rather it estimated.
- **A loopback gateway is not added to the egress allowlist** (`normalizeAllowlistHost` drops it). Deliberate — a container sandbox has its own loopback and could never have reached the host's gateway — and documented: use an in-process backend, or give the gateway a resolvable name.

## Verification

`pnpm turbo run typecheck test build` — 25/25 green. New coverage: registry resolution + validation (`tests/providers.test.ts`), the config/env/egress/sandbox-payload wiring end to end (`tests/config/provider-endpoints.test.ts`), the three `llm.ts` routes including the OpenAI fallback, agentic-pi's parse/register/resolve and its `--providers` flag, and the wizard's overlay output.

Docs synced in the same commit: `spec/02-configuration.md` (new "Provider endpoints" section + env rows + the config shape), `spec/09-sandbox.md`, `config/default.yaml`, the website's configuration page and homepage, and the four affected `CLAUDE.md` files + agentic-pi's README.


cliftonc:
Addressed — pushed 9476012. Two of the three I fixed as asked; the first I fixed differently, and the second turned out to have a sibling defect.

**1. Hard-coded env var name — fixed, but the other way round.** Importing the server's constant into the pi SDK isn't available to us: agentic-pi is a self-contained leaf with no workspace deps (it publishes independently to npm), and core depends on *it*, not the reverse. The name is agentic-pi's own CLI contract anyway — it's the `--providers` fallback — so the constant now lives there and `provider-registry.ts` re-exports it.

It imports `agentic-pi/dist/providers.js` rather than the `agentic-pi` barrel, deliberately: `provider-registry.ts` loads at config time in **every** process, and importing that barrel transitively replaces the global undici dispatcher — the exact hazard `sandbox.ts` uses a lazy `await import()` to avoid. `dist/providers.js` imports nothing at all. (Deep path rather than an `exports` map because adding one would break `apps/evals/src/mechanism.test.ts`, which already deep-imports the package.)

**2. `api` ignored in chat-runner — fixed by refusing the field, not by spreading it.** Spreading `endpoint.api` would have fixed chat and left the sandbox still wrong: `registerProviderOverrides` also drops `api` for a known provider, and it has to — pi's `applyExtension` composes an endpoint override over the built-in catalog by re-pointing those models' `baseUrl` only. Expressing "these models, a different API family" requires enumerating every model, which is exactly what a custom provider entry already is. So a three-way patch would have been three ways to be inconsistent.

`resolveProviderRegistry` now throws when a built-in override changes `api`, naming the alternative ("declare a provider of your own"). All four paths agree by construction, it's a boot-time error instead of a silent divergence, and no capability is lost — an OpenAI-dialect gateway fronting Claude models was always better expressed as its own prefix. Restating a built-in's own dialect stays a no-op.

**3. `envKey` had the identical defect** (found while fixing #2, not flagged in the review): honoured by `llm.ts`, ignored by chat and the sandbox, which both kept resolving the vendor's key. Unlike `api` it *is* expressible everywhere — pi's `configuredApiKey` lets an extension's `apiKey` win — and it's the "key custody" half of why anyone wants a gateway, so I made it work rather than banning it.

The subtlety is that it must not always be passed: `ProviderEndpoint` now carries `envKeyOverridden`, and the key is handed to the model call explicitly **only** when the deployment named one. When it didn't, pi keeps resolving the credential itself — which is what keeps an OAuth subscription login (Claude Pro/Max, Codex, Copilot) working behind a moved endpoint. That negative case has its own test, since it's the one that would break silently.

**4. chat-runner tests** — `resolveModel` and `endpointApiKey` are now exported for tests (the `llm.ts` pattern), with 9 cases: the re-pointed known model, the synthesized custom model including its zero cost rates and declared dialect, both key-lookup positives, the load-bearing negative above, and the unchanged failure modes.

**5. `LASTLIGHT_PROVIDERS` vs `<PREFIX>_BASE_URL` precedence — pinned, behaviour unchanged.** JSON-on-top is the same shape as `LASTLIGHT_MODEL` vs `LASTLIGHT_MODELS` a few lines up in the same resolver, where the JSON map also wins. I'd rather the map behave like its sibling than have one block invent a "more specific name wins" rule, so I documented it in the spec (it was only in the PR body) and added two tests: the conflict case, and the per-key merge when the two name different providers.

**6. contextWindow / maxTokens** — defaults, not ceilings, as you suspected. Renamed `DEFAULT_CUSTOM_MODEL_*` with the reasoning in the comment: the gateway is the only authority on its own limits, and clamping an operator's larger number to ours would break a legitimately bigger context while the gateway would reject an over-long request anyway — with an error about its real limit rather than ours.

Docs updated for the two behaviour changes (`spec/02-configuration.md`, `config/default.yaml`, the website config page). `pnpm turbo run typecheck test build`: 25/25.

